Issue #67, opened 2018-06-08
AsyncWrapper should accept a controller scale


Octogonapus:
`AsyncWrapper` needs to scale the controller response. By default, it can scale it by `127`, but the scale should be user-configurable.

PR #70, opened 2018-06-09
Issue #1

### LINKED ISSUE #1: Add feed forward term to VelPID

### LINKED ISSUE #68: Filter and Controller unit tests should get getOutput()
In addition to checking the result of `filter()`, these tests should also check `getOutput()`.

Octogonapus:
### Description of the Change

Add a feed-forward gain to `IterativeVelPIDController`. Fix bugs in `MedianFilter`, `ComposableFilter` where `getOutput()` did not return the most recent output. Add utility for testing filters and add a test for the new feed-forward response.

### Applicable Issues

Closes #1. Closes #68.

This adds an internal scale to `AsyncWrapper` which should be exposed in #67.


Octogonapus:
Docs are here: https://github.com/purduesigbots/pros-docs/commit/f5d04af5cf85e9c24012853b75528050119c24b9

baylessj:
Tested the changes with the following code:
```
using namespace okapi;
  auto mtr = 1_mtr;
  AsyncVelPIDController con (std::make_shared<IntegratedEncoder>(mtr),
                            std::make_shared<Motor>(mtr), 0.0, 0, 0.05);
  con.setTarget(10);
  // testMtr.move(10000);
  while (true) {
    pros::delay(5);
  }
```
and it works great. Still needs header docstring updates though.

baylessj (approved):
Test code is in the comment above this. Docs updates will be done in pros-docs for now, no need to update the headers' docstrings.

PR #73, opened 2018-06-12
Issue #67: Add AsyncWrapper scale parameter

### LINKED ISSUE #67: AsyncWrapper should accept a controller scale

Octogonapus:
### Description of the Change

Add ctor param for the controller scale with a default value of `127`.

### Benefits

Users can scale their controller output into a different range than `+-127` if they wish.

### Verification Process

This change is simple enough that I did not verify it.

### Applicable Issues

Closes #67.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/ba1abf7be29903caacef7a5346cc17280304c7b6

PR #74, opened 2018-06-12
Fix Issue #72

### LINKED ISSUE #72: ChassisController moveDistance and turnAngle should still operate in units of motor ticks after chassis scales have been set
If a user programs their robot without any scales using the method overloads that take a unitless number (as opposed to those that take a `QLength` or `QAngle`), and then decides to put real scales in, the methods will assume the input number is meters instead of motor ticks (causing unwanted behavior). These method overloads should always operate in motor tick units. If people put in real scales, they can use the overloads that take proper units.

Octogonapus:
### Description of the Change

Divide the input to `moveDistance` and `turnAngle` by their respective scales so they get converted into motor degree units instead of meters.

### Benefits

Programming the robot using `moveDistance` and `turnAngle` without scales, then adding scales and re-running the program will not cause the robot to explode (typically thought of as unexpected behavior).

### Possible Drawbacks

None.

### Verification Process

Tested using:
```
  ChassisControllerIntegrated chassis(-11, 1);
  chassis.moveDistance(200);
  chassis.turnAngle(20);

  ChassisControllerIntegrated chassis2(-11, 1, pros::c::E_MOTOR_GEARSET_36, {2.5_in, 10.5_in});
  chassis2.moveDistance(200);
  chassis2.turnAngle(20);
```

The robot drove the same amount with `chassis` and `chassis2`.

### Applicable Issues

Closes #72.

### DIFF
--- a/include/okapi/chassis/controller/chassisController.hpp
+++ b/include/okapi/chassis/controller/chassisController.hpp
@@ -38,7 +38,7 @@ class ChassisController {
   /**
    * Drives the robot straight for a distance (using closed-loop control).
    *
-   * @param itarget distance to travel in meters
+   * @param itarget distance to travel in motor degrees
    */
   virtual void moveDistance(const double itarget) = 0;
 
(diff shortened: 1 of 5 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/1651d7f2d5673dee2b18598a1feb6f02966b4e2c

PR #78, opened 2018-06-13
Issue #75: Add AbstractTimer and expose millis

### LINKED ISSUE #75: Add AbstractTimer
`Timer` needs to implement some interface so it can be properly mocked. Also, `Timer::millis()` should be exposed so those calls can be mocked.

Octogonapus:
### Description of the Change

Add an interface capturing the methods of `Timer` called `AbstractTimer`. Expose `Timer::millis()` as another method of `AbstractTimer`.

### Benefits

It will be easier to mock `Timer`, and `AbstractTimer` can be used in place of calls to the kernel, so more tests can be moved off the V5 Brain.

### Possible Drawbacks

Another level of indirection to the kernel. Another constructor parameter, though users won't have to deal with it and I plan on more explicitly following the facade pattern in the future.

### Verification Process

The tests still pass.

### Applicable Issues

Closes #75.

### DIFF
--- a/include/okapi/api/util/abstractTimer.hpp
+++ b/include/okapi/api/util/abstractTimer.hpp
@@ -5,75 +5,77 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
-#ifndef _OKAPI_TIMER_HPP_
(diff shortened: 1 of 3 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/f7b1e2d519cbc8345aa8c03d2c44c15a16d0f860

PR #80, opened 2018-06-17
Chassis Model/Controller Factories

Octogonapus:
### Description of the Change

Add a `ChassisModel` and a `ChassisController` factory to decouple the dependency on the PROS kernel from the `ChassisModel` and `ChassisController` interfaces/implementations. Remove simpler constructors from these interfaces and move them into the factories.

### Benefits

No dependencies on the PROS kernel in these interfaces means that they can be tested off of the V5. This moves us closer to moving all the current unit tests off of the V5.

### Possible Drawbacks

Slightly increased API complexity. Users will now have to do
```
  auto chassis = ChassisControllerFactory::create(
    {19, 20}, {-14}, AbstractMotor::motorGearset::E_MOTOR_GEARSET_36, {4_in, 11.5_in});
```
instead of 
```
  ChassisControllerIntegrated chassis(
    {19, 20}, {-14}, AbstractMotor::motorGearset::E_MOTOR_GEARSET_36, {4_in, 11.5_in});
```

### Verification Process

A simple test program like the following confirmed functionality:
```
  auto robot = ChassisControllerFactory::create(
    -11, 1, AbstractMotor::motorGearset::E_MOTOR_GEARSET_36, {2.75_in, 10.5_in});
  robot.moveDistance(100);
  robot.moveDistance(3_in);
```

### DIFF
--- a/include/okapi/api.hpp
+++ b/include/okapi/api.hpp
@@ -12,6 +12,7 @@
 #include "okapi/api/chassis/controller/chassisControllerIntegrated.hpp"
 #include "okapi/api/chassis/controller/chassisControllerPid.hpp"
 #include "okapi/api/chassis/controller/chassisScales.hpp"
+#include "okapi/api/chassis/model/chassisModelFactory.hpp"
 #include "okapi/api/chassis/model/skidSteerModel.hpp"
 #include "okapi/api/chassis/model/threeEncoderSkidSteerModel.hpp"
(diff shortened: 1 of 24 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/cfc9857a1159962ddc810fce89210de1835b62cc

PR #87, opened 2018-06-23
Spike: Standalone API

Octogonapus:
### Description of the Change

This is a more complete separation than the last similar PR. This is a full separation of PROS kernel independent/dependent classes into api/impl, respectively. Files in api have been added to a cmake build and all their tests have been moved over as well.

### Benefits

~140 of the tests have been moved off the V5. These can now be run as part of a CI build (though this PR does not add that functionality to `.travis.yml`). Additionally, these files integrate well with a modern IDE.

### Possible Drawbacks

Slightly increased API complexity. I have introduced a `VelMathFactory` which users should use instead of the `VelMath` constructor. Otherwise, the API remains unchanged.

### Verification Process

Everything still compiles and the unit tests still pass.

### Applicable Issues

I did not create an issue for this spike.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/85ee40801d76d2afdd9defec5a8204acbf38918d

PR #96, opened 2018-07-09
Issue #95: Add implementation of Supplier

### LINKED ISSUE #95: Add generic Supplier
The "supplier pattern" is starting to find use with the new factories. Instead of hiding the pattern behind `std::function<Foo(void)>`, there should be a templated `Supplier` class that better shows the intent.

Example code that works on cpp.sh:
```c++
#include <iostream>
#include <string>
#include <functional>

template<typename T> class Supplier {
    public:
    explicit Supplier(std::function<T(void)> ifunc): func(ifunc) {}
    virtual ~Supplier() = default;
    
    T get() const {
        return func();
    }
    
    protected:
    std::function<T(void)> func;
};

int main()
{
    Supplier<std::string> stringSupplier([]() { return "Hello, world!"; });
    std::cout << stringSupplier.get() << std::endl;
    
    Supplier<int> intSupplier([]() { return 42; });
    std::cout << intSupplier.get() << std::endl;
}
```

Octogonapus:
### Description of the Change

This PR adds a `Supplier` template class that accepts a `std::function` and serves to clarify the intent of the function. `Supplier` is meant to be used when multiple instances of one type are needed.

### Benefits

A simpler and more readable API.

### Possible Drawbacks

None.

### Verification Process

This simple change was not verified.

### Applicable Issues

Closes #95.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,4 +118,4 @@ add_executable(OkapiLibV5
         test/crossPlatformTestRunner.cpp
         test/filterTests.cpp
         test/testMain.cpp
-        test/utilTests.cpp include/test/tests/api/implMocks.hpp test/implMocks.cpp)
+        test/utilTests.cpp include/test/tests/api/implMocks.hpp test/implMocks.cpp include/okapi/api/util/supplier.hpp)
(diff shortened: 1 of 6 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/09ba27b4e593e7b515c3b4c0a7359c779380d0bb

PR #100, opened 2018-07-11
Issue #94: Move ChassisControllers into API

### LINKED ISSUE #94: Move ChassisControllers into api
With `PIDTuner` already moved in another branch, these are the last two non-factory and non-device type files left in `impl` that could be moved into `api`. We just need to inject an `AbstractTimer` supplier and a `SettledUtil` supplier.

Octogonapus:
### Description of the Change

This PR moves `ChassisControllerIntegrated` and `ChassisControllerPID` into `api`.

### Benefits

With the exception of `PIDTuner` which has been moved on another branch, the only things left in `impl` are factories and classes that interface directly with the kernel! :tada: 

### Possible Drawbacks

Slightly increased API complexity for advanced users.

### Verification Process

This change was not verified.

### Applicable Issues

Closes #94.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/fbc6b3d90b14791e2ddb4f2372bfe3ae9c86f5b8

PR #103, opened 2018-07-13
Issue #102: Add TimeUtil

### LINKED ISSUE #102: Create a class to hold AbstractTimer, AbstractRate, SettledUtil
These three classes are often passed together and as a result, increase the number of parameters for many constructors. They are used together frequently enough that there should be a simple class to hold all three of them that can be passed in instead. Additionally, there should also be a factory for creating this class.

Octogonapus:
### Description of the Change

A utility class was added to simplify the constructors for classes which need time-related classes such as `AbstractTimer`, `AbstractRate`, and `SettledUtil`.

### Benefits

Greatly simplified constructor calls.

### Possible Drawbacks

None.

### Verification Process

The tests still pass.

### Applicable Issues

Closes #102.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,7 @@ add_executable(OkapiLibV5
         test/filterTests.cpp
         test/utilTests.cpp
         include/test/tests/api/implMocks.hpp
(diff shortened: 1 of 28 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/841df3b40fc31b0ad0afd1faf1a03f64a6fec9a2

PR #109, opened 2018-07-17
Issue #5: PID Tuner Polishing

### LINKED ISSUE #5: libblrs PID autotuner
https://github.com/purduesigbots/libblrs/blob/master/libfbc/src/fbc_util.c

baylessj:
### Description of the Change

PID Autotuner now works like a charm on any generic PID controller.  

### Benefits

PID Autotuning capability for any generic PID controller

### Possible Drawbacks

Not quite the same drop in usage with a `ChassisModel` or whatever it was before, but this should be more extensible in the future.

### Verification Process

Tested with a single V5 Motor and the follow code at the top of opcontrol:

```
pros::c::lcd_initialize();
  std::shared_ptr<AbstractMotor> testMotor = std::make_shared<okapi::Motor>(1);
  auto timer = 5_s;
  PIDTuner tuner = PIDTunerFactory::create(testMotor->getEncoder(), testMotor, timer, 2000, 0.0001,
                        0.002, 0.0, 1.0, 0.2, 0.4, 2);
  IterativePosPIDControllerArgs args = tuner.autotune();
  pros::c::lcd_print(0, "%f %f %f\n", args.kP, args.kI, args.kD);
```

### Applicable Issues

Closes #5


include/main.h
```diff
@@ -33,12 +33,11 @@
  *            The okapi namespace will export all symbols inside the pros namespace.
  */
 // using namespace pros;
-// using namespace okapi;
 
 /**
  * You should add more #includes here
  */
-//#include "okapilib.h"
+// #include "okapi/api.hpp"
```
Octogonapus: This should be changed in the PROS template as well.

src/api/control/iterative/iterativePosPidController.cpp
```diff
@@ -10,6 +10,7 @@
 #include "okapi/api/control/iterative/iterativePosPidController.hpp"
 #include <algorithm>
 #include <cmath>
+#include "api.h"
```
Octogonapus: You can't include that file inside of `api`.

src/api/control/iterative/iterativePosPidController.cpp
```diff
@@ -106,6 +107,7 @@ double IterativePosPIDController::step(const double inewReading) {
     if (loopDtTimer->getDtFromHardMark() >= sampleTime) {
       error = target - inewReading;
 
+
```
Octogonapus: I'm not sure if `clang-format`  will remove this duplicate newline (go ahead and do it manually if it doesn't), but make sure you run it before merging.

src/api/control/util/pidTuner.cpp
```diff
@@ -0,0 +1,172 @@
+/**
+ * @author Jonathan Bayless, Team BLRS
+ * @author Ryan Benasutti, WPI
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+#include "okapi/api/control/util/pidTuner.hpp"
+#include "api.h"
```
Octogonapus: You can't include that file inside of `api`.

src/api/control/util/pidTuner.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+  ParticleSet global{};
+  global.kP.best = 0;
+  global.kI.best = 0;
+  global.kD.best = 0;
+  global.bestError = std::numeric_limits<double>::max();
+
+  // Run the optimization
+  for (int iteration = 0; iteration < numIterations; iteration++) {
+    // Test constants then calculate fitness function
```
Octogonapus: I don't think this comment makes sense here. Maybe it needs to be moved down a bit?

src/api/control/util/pidTuner.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+      // Reverse the goal every iteration to stay in the same general area
+      std::int32_t target = goal;
+      if (!firstGoal) {
+        target *= -1;
+      }
+      firstGoal = !firstGoal;
+
+      testController.setTarget(target);
+      timeUtil.getSettledUtil()->reset();
```
Octogonapus: `getSettledUtil()` returns a _new_ `SettledUtil`. Resetting it won't help you. You need to keep the same one around and reset _that_ one.

Octogonapus (changes requested):

include/main.h
baylessj: That fix is on its way with 3.1.0

src/api/control/iterative/iterativePosPidController.cpp
baylessj: That was just for testing, I'll take that out

src/api/control/util/pidTuner.cpp
baylessj: Going to remove this since the settled util is already used properly by the iterative controller

include/okapi/api/control/util/pidTuner.hpp
```diff
@@ -10,43 +10,50 @@
...
   virtual IterativePosPIDControllerArgs autotune();
 
   protected:
-  static constexpr double inertia = 0.5;   // Particle intertia
+  static constexpr double inertia = 0.5;   // Particle inertia
   static constexpr double confSelf = 1.1;  // Self confidence
   static constexpr double confSwarm = 1.2; // Particle swarm confidence
   static constexpr int increment = 5;
   static constexpr int divisor = 5;
+  static constexpr QTime loopDelta = 100_ms; // NOLINT
```
Octogonapus: `100_ms` seems long to me. Did it work better than `10_ms`? Have you experimented with values between those two?

I remember motors sometimes missing updates when I queried them every `10_ms` so I believe that `100_ms` might be more stable.

baylessj: ah rip that was also another holdover from testing, thank you for catching that. The motors accept changes every 10ms as per their software, I'll update the loopDelta accordingly.

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/af22de266785b30ff760dcb45357dde92e353ecf

Octogonapus:
If you want, you are welcome to add an explanation of the tuning algorithm to those docs.

baylessj:
Docs updated to purduesigbots/pros-docs@93139e1

Octogonapus (approved):

PR #112, opened 2018-07-20
Issue #101: Add ReadOnlyChassisModel

### LINKED ISSUE #101: Make ReadOnlyChassisModel
With CM now accepted as a `unique_ptr` in the CC ctor, the odom flavor CC's break down because they need to share the CM. There should be a "read-only" version which will support `shared_ptr` so these other classes can use it, while still enforcing the one-writer policy. Additionally, the current CM's copy ctor and copy-assignment operator should be deleted.

`unique_ptr` accomplishes this by doing,
```c++
unique_ptr(const unique_ptr&) = delete;
unique_ptr& operator=(const unique_ptr&) = delete;
```

Octogonapus:
### Description of the Change

There should be a "read-only" `ChassisModel` to help enforce a one-writer policy. This PR introduces that interface.

### Benefits

Classes can now accept a `ChassisModel` without owning it.

### Possible Drawbacks

None.

### Verification Process

This pure API change was not verified.

### Applicable Issues

Closes #101.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ add_executable(OkapiLibV5
         test/unitTests.cpp
         test/implMocks.cpp
         src/api/util/timeUtil.cpp
-        include/okapi/api/util/timeUtil.hpp)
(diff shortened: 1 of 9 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/a0712f4f470d4b65b9f1f1330bb4111517406669

PR #114, opened 2018-07-21
Issue #110: Support higher-dimensional controllers

### LINKED ISSUE #110: Support higher-dimensional controllers
With 2D path planning in the near future, OkapiLib will have to support closed-loop controllers with higher input dimensions than 1, and potentially different input types than `double`. One way this could be solved is by templating the parameter to `step()`. In this way, implementing classes would specify the type of their input. This also satisfies the requirement of higher-dimensional input because implementing classes could define their own input class. Existing classes would specify this template parameter as `double` so the API would remain unchanged.

Octogonapus:
### Description of the Change

This PR adds some simple architecture to support higher-dimensional controllers through the use of a template parameter for the input to the controller. Controllers will then be able to define their own input type, instead of using a `double` like the code works currently.

### Benefits

Outlined above.

### Possible Drawbacks

None.

### Verification Process

I made a small test implementation of a controller with a custom input type to confirm the template parameter worked like I wanted it to.

### Applicable Issues

Closes #110.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/e7ae3b56a302fce88f778f36af3131ebd642fab8

PR #118, opened 2018-07-26
Issue #84: Async Chassis Control

### LINKED ISSUE #84: Async ChassisController
Add something to run a `ChassisController` async.

baylessj:
### Description of the Change

Add Asynchronous movement functionality to the `ChassisController`s. Can set a target and then call `waitUntilSettled()` when it is desired to wait for the movement to complete.

### Benefits

Can operate a `ChassisController` at the same time as other subsystems. Generally more freedom with movement timing.

### Possible Drawbacks

No intelligent prioritization of movements - if `moveDistanceAsync` is called from two different threads, then the most recent call will always take priority. There is also no protection in the `waitUntilSettled()` function to prevent the goal from being changed.

This is intentional, however, as it was decided that it would be best to prioritize having the ability to override a `waitUntilSettled()` call with, for instance, an emergency stop from a different thread.

### Verification Process

- Wrote Tests
- Run a chassis controller alongside another subsystem
Test Code:
```
  auto chassis =
    ChassisControllerFactory::create({-11, -20}, {10}, AbstractMotor::gearset::red, {4_in, 11.5_in});
  auto asyncMotor = AsyncControllerFactory::posIntegrated(2);

  chassis.moveDistanceAsync(1_m);
  asyncMotor.setTarget(1000);
  chassis.waitUntilSettled();
```

### Applicable Issues

Closes #84

### DIFF
--- a/src/api/chassis/controller/chassisControllerPid.cpp
+++ b/src/api/chassis/controller/chassisControllerPid.cpp
@@ -53,8 +53,6 @@ void ChassisControllerPID::loop() {
   double distanceElapsed = 0, angleChange = 0;
   modeType pastMode = distance;
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wmissing-noreturn"
   while (!dtorCalled) {
     /**
      * doneLooping is set to false by moveDistanceAsync and turnAngleAsync and then set to true by
@@ -63,7 +61,6 @@ void ChassisControllerPID::loop() {
     if (!doneLooping) {
       if (mode != pastMode) {
         encStartVals = model->getSensorVals();
-        encVals = 0;
 
         distancePid->reset();
         anglePid->reset();
@@ -93,7 +90,6 @@ void ChassisControllerPID::loop() {
 
     rate->delayUntil(10_ms);
   }
-#pragma clang diagnostic pop
 }
 
 void ChassisControllerPID::trampoline(void *context) {
@@ -110,6 +106,8 @@ void ChassisControllerPID::moveDistanceAsync(const QLength itarget) {
   const double newTarget = itarget.convert(meter) * straightScale * gearRatio;
   distancePid->setTarget(newTarget);
   anglePid->setTarget(0);
+
+  doneLooping = false;
 }
 
 void ChassisControllerPID::moveDistanceAsync(const double itarget) {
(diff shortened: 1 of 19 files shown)


src/api/chassis/controller/chassisControllerPid.cpp
```diff
@@ -42,62 +43,132 @@ ChassisControllerPID::ChassisControllerPID(
...
 void ChassisControllerPID::turnAngle(const double idegTarget) {
   // Divide by turnScale so the final result turns back into motor degrees
   turnAngle((idegTarget / turnScale) * degree);
 }
+
+void ChassisControllerPID::waitUntilSettled() {
+  switch (mode) {
+    case distance:
+      while (!distancePid->isSettled() && !anglePid->isSettled()) {
+        rate->delayUntil(10_ms);
```
Octogonapus: I think there should be a check in these loops the monitor `mode`. What if, while waiting for a `moveDistance` to settle, the user calls `turnAngle`? I think we will still be waiting for the first call to settle.

test/chassisControllerTests.cpp
```diff
@@ -90,6 +90,46 @@ TEST_F(ChassisControllerIntegratedTest, MoveDistanceUnitsTest) {
   assertMotorsHaveBeenStopped(leftMotor, rightMotor);
 }
 
+TEST_F(ChassisControllerIntegratedTest, MoveDistanceAsyncRawUnitsTest) {
+  controller->moveDistanceAsync(100);
+
+  EXPECT_TRUE(leftController->disabled);
+  EXPECT_TRUE(rightController->disabled);
```
Octogonapus: What are you trying to test here? After you call `moveDistanceAsync`, the controllers should actually be enabled while the robot is moving. If this passes, it's because the call to `moveDistanceAsync` fully completes before code execution moves to the assertions. The same goes for the rest of these tests.

Octogonapus (changes requested):
I also want to see tests for calling `moveDistance` and then `turnAngle` before `moveDistance` has settled, and vice versa. This will probably require an implementation of `SettledUtil` that never settles (or if you wanted to go the extra mile, an implementation that lets you toggle whether it's settled).

Octogonapus:
If you want this PR to close #84, you need to say "closes" #84, "fixes" #84, etc. You can't have another word in between the keyword and the issue number.

test/chassisControllerTests.cpp
baylessj: This is the result of some poor "find and replace" usage to remove usage of `Controller->enabled`, which obviously didn't compile. I'll take a closer look over everything to make sure that it's `EXPECT_FALSE` where necessary

Octogonapus (approved):

Octogonapus:
@baylessj Please do not merge a PR that has the `needs docs` label without writing + linking the docs. Additionally, I wanted you to test this with real hardware before merging.

baylessj:
Ah my bad, I saw the approval and figured it was good to go. I'll take care of the docs tonight.

Octogonapus:
@baylessj I am currently doing the docs.

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/67853df86d183abfb21bd285024ba32b01ac1ff8

baylessj:
Well that works too lol. Thanks!

PR #125, opened 2018-07-31
Issue #123: Make isSettled return true if the controller is disabled

### LINKED ISSUE #123: ClosedLoopController::isSettled should return true if the controller is disabled
In order to prevent program hangs when waiting for a disabled controller to settle, `ClosedLoopController::isSettled` should return `true` if the controller is disabled.

Octogonapus:
### Description of the Change

In order to prevent program hangs when waiting for a disabled controller to settle, `ClosedLoopController::isSettled` should return `true` if the controller is disabled.

### Benefits

Described above.

### Possible Drawbacks

If the controller is moved by external forces while disabled, then `isSettled()` will not reflect that. `getError()` would still show that movement, though. This could have a small impact on some users.

### Verification Process

The tests still pass :)

### Applicable Issues

Closes #123.

### DIFF
--- a/include/okapi/api/control/async/asyncPosIntegratedController.hpp
+++ b/include/okapi/api/control/async/asyncPosIntegratedController.hpp
@@ -45,6 +45,8 @@ class AsyncPosIntegratedController : public AsyncPositionController {
    * Returns whether the controller has settled at the target. Determining what settling means is
    * implementation-dependent.
    *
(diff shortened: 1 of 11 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/89a4e331a80f50094e90767623c7898132458003

PR #126, opened 2018-07-31
Issue #122: Make reset non-const

### LINKED ISSUE #122: `ContinuousRotarySensor::reset()` should not be const
`ContinuousRotarySensor::reset()` can sometimes involve calls to software instead of just resetting hardware, so this method should not be `const`.

Octogonapus:
### Description of the Change

`ContinuousRotarySensor::reset()` can sometimes involve calls to software instead of just resetting hardware, so this method should not be `const`.

### Applicable Issues

Closes #122.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/a621bf64812b15c284814e29ec2f67db30de51b0

PR #131, opened 2018-08-01
Issue #129: Override stop in the ChassisControllers

### LINKED ISSUE #129: ChassisControllerPid should override stop
`ChassisControllerPid` should override `stop()` so that it also properly stops the internal PID task and disables the controllers. If someone calls `stop()` while a movement is in progress, the PID task may override the call to `stop()`.

Octogonapus:
### Description of the Change

The `ChassisController` implementations need to override `stop()` so they can disable their internal controllers, stop PID loops, etc. `stop()` was also made non-const.

### Verification Process

This change was tested with CCPID on a real robot, and some tests were written to assert that motors get stopped and controllers get disabled.

### Applicable Issues

Closes #129.

### DIFF
--- a/include/okapi/api/chassis/controller/chassisController.hpp
+++ b/include/okapi/api/chassis/controller/chassisController.hpp
@@ -118,7 +118,7 @@ class ChassisController {
   /**
    * Stop the robot (set all the motors to 0).
    */
-  virtual void stop() const;
+  virtual void stop();
 
   /**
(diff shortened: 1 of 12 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/4353eb1a28a4bcbcd223357e93580d45c4d7277b

PR #136, opened 2018-08-03
Issue #135: Add logging support

### LINKED ISSUE #135: Logging Support
OkapiLib needs logging support for the debugging purposes of both users and us as developers trying to help users. I want this to be a no-overhead setup, though. If a user isn't using a log level, they shouldn't have to pay for it in cpu cycles, program size, etc. There should also be support for selecting the stream to output on. By default, for now, this can be `stdout` which is `/ser/sout`. I have this setup as the `LOG_STREAM` define.

Octogonapus:
### Description of the Change

This PR adds a logger class so OkapiLib's classes can log information that may help users when they are debugging their programs. Various log levels (including `off`) are supported, and the output file is configurable. Logging is disabled by default.

### Benefits

Users will have an easier time debugging OkapiLib's behavior and we will have an easier time helping them.

### Possible Drawbacks

At high log levels, it's possible that programs may slow down such that they start missing some deadlines (i.e., in tight loops). I've tested two tasks logging 70 characters each every 1 ms with no problems, but that is essentially a microbenchmark and its applicability to real programs is limited. OkapiLib's classes should be very respectful and thoughtful of the logging levels, what gets logged, and how much gets logged.

### Verification Process

This was tested on a V5 and on my desktop. On the V5, I output to `stdout` and on my desktop I output to a file in a test.

### Applicable Issues

Closes #135.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,7 @@ add_executable(OkapiLibV5
         test/unitTests.cpp
         test/implMocks.cpp
         src/api/util/timeUtil.cpp
-        include/okapi/api/util/timeUtil.hpp include/okapi/api/chassis/model/readOnlyChassisModel.hpp src/api/chassis/model/readOnlyChassisModel.cpp)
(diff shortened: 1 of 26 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/a919e0d4ab066774be7852d3dd50419b9b5a8a08

PR #138, opened 2018-08-04
Add Logger tests

Octogonapus:
### Description of the Change

When I made the Logger PR I didn't think it would be reasonable to test it; however, I just learned that you can open an in-memory file using `open_memstream`, so here are some tests using that. Certified leak-free using Valgrind :)

### Applicable Issues

None.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/d631eea4a01527340420975db30eff6ec3fa66bb

PR #143, opened 2018-08-05
Issue #137: Add 2D Motion Profiling using Pathfinder

### LINKED ISSUE #137: 2D Motion Profiling
I'd like to have some kind of 2D motion profiling on release, so it's worth investigating if the old verison of Pathfinder is usable for our purposes: https://github.com/JacisNonsense/Pathfinder/tree/master/Pathfinder-Core
This would get upgraded as soon as the next version is released.

Octogonapus:
### Description of the Change

This PR adds 2D motion profiling using Pathfinder. It also rewrote some of the structure around closed-loop controllers so they are templated all the way through, paving the way for both this PR and Pure Pursuit.

### Benefits

Motion profiling of course :)

### Possible Drawbacks

None.

### Verification Process

There are a few new tests, and this has been tested on a real robot.

### Applicable Issues

Closes #137.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/12aee5a8d2a06702f977fb54f4f5bc184dc8fafc

PR #145, opened 2018-08-06
Issue #141: Add AsyncController factory methods for pots

### LINKED ISSUE #141: Add Controller Factory methods that take a potentiometer
If users want to put a pot on, for example, their lift, then they will have to use an overload that accepts a `std::shared_ptr`. There should be overloads which accept a potentiometer directly.

Octogonapus:
### Description of the Change

This PR adds methods to the `AsyncControllerFactory` which take a `Potentiometer`.

### Verification Process

This was not verified.

### Applicable Issues

Closes #141.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/499fae58edae03c1878a7fdf4b22272be34098ab

PR #146, opened 2018-08-06
Issue #142: Add another PID controller for turns

### LINKED ISSUE #142: Add separate PID controller for turning in CCPID
CCPID uses the same PID controller for turning as it does for driving straight. There should be a separate one for turning so users can have different gains.

Octogonapus:
### Description of the Change

This PR adds a third PID controller to `ChassisControllerPID` so there are separate PID controllers for turns and for angle correction. By default, the turn PID controller and the angle correction PID controller will share the same set of gains, as was the case before this change.

### Benefits

Often times having one set of gains for both turns and angle correction does not work optimally in both cases. This change lets users have a different set of gains for each case.

### Possible Drawbacks

None.

### Verification Process

The existing tests were extended to test the new controller.

### Applicable Issues

Closes #142.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/92ba1233a6b20c44d85e56d1a7f3a8cf2a6c4bf5

PR #152, opened 2018-08-07
Issue #149: Change default error sum bounds

### LINKED ISSUE #149: Change error sum bounds
Lachlan pointed out to me that the current error sum bounds make no sense. They should be `[0, inf]` or something like that so the controller always integrates, and users should change them if they want better performance.

Octogonapus:
### Description of the Change

Lachlan pointed out to me that the current error sum bounds of `(500, 1250)` make no sense. For the record, these came out of a conversation I had with Jason and made sense for the Cortex platform, but they aren't relevant anymore for V5. The new bounds are `(0, std::numeric_limits<double>::max())`  so the controller always integrates. Users can them if they want better performance.

### Benefits

The integrator should behave closer to what people expect.

### Possible Drawbacks

None.

### Verification Process

Some new tests were added for the integrator bounds.

### Applicable Issues

Closes #149.

### DIFF
--- a/include/okapi/api/control/iterative/iterativePosPidController.hpp
+++ b/include/okapi/api/control/iterative/iterativePosPidController.hpp
@@ -14,6 +14,7 @@
 #include "okapi/api/control/util/settledUtil.hpp"
 #include "okapi/api/util/logging.hpp"
 #include "okapi/api/util/timeUtil.hpp"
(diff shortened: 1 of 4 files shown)


Octogonapus:
Docs: Docs for OkapiLibV5 PR 152

PR #160, opened 2018-08-12
Clean up some warnings and unused args classes

Octogonapus:
### Description of the Change

This PR does some housekeeping to cut down on a few warnings and remove the unused "args" classes before the release of v3.1.0.

### Verification Process

The tests still pass :)

### Applicable Issues

None.


src/api/control/util/pidTuner.cpp
```diff
@@ -105,7 +105,7 @@ PIDTuner::Output PIDTuner::autotune() {
       const double start_val = input->controllerGet();
 
       QTime settleTime = 0_ms;
-      int itae = 0;
+      double itae = 0;
```
Octogonapus, line 108: @baylessj It's fine for this to be a `double` right? I noticed there was a narrowing conversion to an `int` so this should give a more accurate result.

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/0d77696ed51216987ded38b041160c453a796c4d

PR #161, opened 2018-08-12
Issue #147: Run memcheck and massif with builds

### LINKED ISSUE #147: Run Valgrind on a Travis CI build
We should start running Valgrind with CI builds to check for memory leaks.
We probably want the options:
`--error-exitcode=1 --show-reachable=no --track-origins=yes --num-callers`

Octogonapus:
### Description of the Change

Valgrind's Memcheck and Massif tools are great for detecting undefined behavior and memory leaks, so we should run them with the CI builds.

### Benefits

Increased confidence in program correctness and stability.

### Possible Drawbacks

Longer builds :(

### Verification Process

N/A

### Applicable Issues

Closes #147.

### DIFF
--- a/include/okapi/api/chassis/controller/chassisControllerPid.hpp
+++ b/include/okapi/api/chassis/controller/chassisControllerPid.hpp
@@ -125,8 +125,8 @@ class ChassisControllerPID : public virtual ChassisController {
   bool waitForAngleSettled();
   void stopAfterSettled();
 
-  typedef enum { distance, angle } modeType;
-  modeType mode;
+  typedef enum { distance, angle, none } modeType;
+  modeType mode{none};
 };
 } // namespace okapi
 
--- a/include/okapi/api/chassis/controller/chassisControllerPid.hpp
+++ b/include/okapi/api/chassis/controller/chassisControllerPid.hpp
@@ -108,8 +108,7 @@ class ChassisControllerPID : public virtual ChassisController {
   void stop() override;
 
   /**
-   * Start the internal thread to loop the PID controllers. This should not be called by normal
-   * users.
+   * Starts the internal thread. This should not be called by normal users.
    */
   void startThread();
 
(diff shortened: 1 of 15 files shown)


include/okapi/api/chassis/controller/chassisControllerPid.hpp
```diff
@@ -105,6 +107,11 @@ class ChassisControllerPID : public virtual ChassisController {
    */
   void stop() override;
 
+  /**
+   * Starts the internal thread. This should not be called by normal users.
```
Octogonapus: Elaborate that this is called by the factories. If users make an instance of this class manually (i.e. without using the factory) then they must call this method before using the class.

include/okapi/impl/control/async/asyncControllerFactory.hpp
```diff
@@ -322,13 +322,13 @@ class AsyncControllerFactory {
    * @param imaxVel The maximum possible velocity.
    * @param imaxAccel The maximum possible acceleration.
    * @param imaxJerk The maximum possible jerk.
-   * @param imodel The chassis model to control.
+   * @param imodel The chassis to control.
```
Octogonapus: This should be `ichassis` not `imodel`.

memcheck-suppressions.txt
```diff
@@ -0,0 +1,7 @@
+{
+   pathfinder_trajectorycandidate_saptr_laptr_leak
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:pathfinder_prepare
+}
```
Octogonapus: @JacisNonsense says that the two fields in there that get leaked can be freed safely after modifying a path for tank drive.

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/a812e4010bd7f97ec87e7bb7b377fe35fe7601b7

PR #175, opened 2018-08-17
Add more IMVC tests

Octogonapus:
Fix some weird edge-case bugs in the iterative controllers around getting error when disabled.

### Description of the Change

Tests :)
This also fixes some weird edge-case bugs in the Iterative controllers where `getError()` would not update properly if the controller was being stepped while disabled.

### Applicable Issues

None.


src/api/control/iterative/iterativeVelPidController.cpp
```diff
@@ -103,7 +106,8 @@ double IterativeVelPIDController::getOutput() const {
 }
 
 double IterativeVelPIDController::getError() const {
-  return error;
+  return target - velMath->getVelocity().convert(rpm);
+  ;
```
Octogonapus: Extra semicolon 🤔

PR #187, opened 2018-08-29
Issue #186: Add VelMath sampleTime

### LINKED ISSUE #186: Add sampleTime parameter to VelMath
`VelMath` should have a `sampleTime` parameter so that users can put a lower bound on the frequency of velocity measurements, similarly to how pos and vel PID work.

Octogonapus:
### Description of the Change

`VelMath` should have a `sampleTime` parameter so that users can put a lower bound on the frequency of velocity measurements, similarly to how pos and vel PID work.

I have also taken the opportunity (since I am modifying `AbstractTimer` and `Timer` with a new method `readDt()`) to move most of the timer-related logic up from `Timer` into `AbstractTimer`. This simplifies the `MockTimer` logic since it can be reused.

### Benefits

More configurable velocity measurements.

### Possible Drawbacks

None.

### Verification Process

Some new tests were added to test the `sampleTime` behavior.

### Applicable Issues

Closes #186.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/f1a9ea528f114acd412cf1adcd64605acc9b6a07

PR #196, opened 2018-09-11
Issue #195: Add createPtr versions of all the CCF methods

### LINKED ISSUE #195: Add pointer-type methods to ChassisControllerFactory
Some users are trying to share a `ChassisController` directly between files (instead of wrapping it in another class) and then, of course, run into the problem that the two existing `ChassisController` implementations are not copyable. A simple fix is to add `createPtr()` methods in addition to the `create()` methods of `ChassisControllerFactory`.

Octogonapus:
### Description of the Change

Some users are trying to share a `ChassisController` directly between files (instead of wrapping it in another class) and then, of course, run into the problem that the two existing `ChassisController` implementations are not copyable. This PR adds`createPtr()` methods in addition to the `create()` methods of `ChassisControllerFactory`.

### Benefits

Sharing a pointer between files is easier for a lot of users compared to writing a "drive" class that can be shared instead.

### Possible Drawbacks

None.

### Verification Process

This was verified by hand.

### Applicable Issues

Closes #195.

### DIFF
--- a/include/okapi/impl/chassis/controller/chassisControllerFactory.hpp
+++ b/include/okapi/impl/chassis/controller/chassisControllerFactory.hpp
@@ -43,11 +43,43 @@ class ChassisControllerFactory {
    * @param igearset motor internal gearset and gear ratio
    * @param iscales see ChassisScales docs
    */
+  static std::shared_ptr<ChassisControllerIntegrated>
+  createPtr(Motor ileftSideMotor,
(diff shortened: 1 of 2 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/7b5a4f8102493a490bec94d6f73811499af03259

PR #207, opened 2018-09-21
1D Motion Profiling

### LINKED ISSUE #198: 1D Motion Profiling with Pathfinder
Pathfinder's "unmodified" profiles are 1D trapezoidal profiles. We should expose this to the user for 1DOF systems such as lifts.

Octogonapus:
### Description of the Change

Pathfinder's "unmodified" profiles are 1D trapezoidal profiles. We should expose this to the user for 1DOF systems such as lifts. This PR adds that, adds a few missing telemetry methods to `AbstractMotor` that the new 1D motion profile class needed (albeit indirectly) and makes some async controllers implement `ControllerOutput`.

Still need to do:
- [x] Investigate (and update) other Async Controllers that could implement `ControllerOutput`.
- [x] Add the new class to the `AsyncControllerFactory`.

### Benefits

1D motion profiling of course :)

### Possible Drawbacks

I'm not thrilled with the lack of code reuse, but I can't think of a good way to split this in a concise way that will make sense to people looking to extend the features.

### Verification Process

Tested by hand. Still need to do:
- [x] Test the most recent changes again.

### Applicable Issues

Closes #198.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ add_executable(OkapiLibV5
         include/okapi/api/chassis/model/threeEncoderSkidSteerModel.hpp
         include/okapi/api/chassis/model/xDriveModel.hpp
         include/okapi/api/control/async/asyncController.hpp
+        include/okapi/api/control/async/asyncLinearMotionProfileController.hpp
         include/okapi/api/control/async/asyncMotionProfileController.hpp
         include/okapi/api/control/async/asyncPosIntegratedController.hpp
         include/okapi/api/control/async/asyncPositionController.hpp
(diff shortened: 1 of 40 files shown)


Octogonapus:
Something weird about the API I have right now is that the waypoints are taken in as `QLength`. This is great for linear lifts, but for linkages, joints, etc. it doesn't make much sense. I'm not sure if we can use units for this, maybe `double` will work the best. @HotelCalifornia or @baylessj do you guys have any input?

HotelCalifornia:
as discussed offline, I think both should be provided (and IIRC @baylessj was of the opinion they should just be doubles)

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/569205f47cce15dc15c1279f2b2667be1126cdec

PR #208, opened 2018-09-21
Throw exception from MotorGroup ctor if no motors are given

### LINKED ISSUE #205: MotorGroup should throw an exception for no motors
`MotorGroup` should throw a `std::invalid_argument` if it isn't given any motors. The class does not perform any bounds checking for speed purposes, so having no motors would invoke undefined behavior.

Octogonapus:
### Description of the Change

`MotorGroup` should throw a `std::invalid_argument` if it isn't given any motors. The class does not perform any bounds checking for speed purposes, so having no motors would invoke undefined behavior.

### Verification Process

This change was not verified; however, I did verify a small test class with the exact same constructor code.

### Applicable Issues

Closes #205.

### DIFF
--- a/include/okapi/impl/device/motor/motorGroup.hpp
+++ b/include/okapi/impl/device/motor/motorGroup.hpp
@@ -16,6 +16,13 @@
 namespace okapi {
 class MotorGroup : public AbstractMotor {
   public:
+  /**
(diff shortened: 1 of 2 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/b7ae3484e2b9abb52b746a678ab0add5893cf894

PR #210, opened 2018-09-21
Fix APIC RPM limit

### LINKED ISSUE #209: APIC limits motor RPM to 127
`AsyncPosIntegratedController` limits the RPM of profiled movements to `127` rpm. It should set the maximum allowable RPM (`600`) and let vexOS limit unless the user wants something else.

Octogonapus:
### Description of the Change

`AsyncPosIntegratedController` hardcoded `127` RPM for profiled movements. This PR fixes that by sending max RPM (`600`) by default and exposing it as a user-configurable parameter. This PR also changes the stop behavior from setting `0` voltage to setting `0` velocity so stopping respects the motor's brake mode.

### Benefits

Bug fix and configurable max velocity :)

### Possible Drawbacks

None.

### Verification Process

Some new tests were added.

### Applicable Issues

Closes #209.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ add_executable(OkapiLibV5
         include/okapi/api/chassis/model/threeEncoderSkidSteerModel.hpp
         include/okapi/api/chassis/model/xDriveModel.hpp
         include/okapi/api/control/async/asyncController.hpp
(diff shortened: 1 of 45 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/26109256684a04cd89a8094c376e889ab34f46a1

PR #220, opened 2018-09-23
ChassisModel max velocity is not always correct

### LINKED ISSUE #219: ChassisModel max velocity is not always correct
`ChassisModel` has a default max velocity of `600` which can be wrong depending on the current gearset. The max velocity should not be a ctor param and should be inferred from the gearset given. If users want to change it specially, there is a method exposed for that (and for voltage).

Octogonapus:
### Description of the Change

`ChassisModel` has a default max velocity of `600` which can be wrong depending on the current gearset. This PR makes the max velocity a required parameter to `ChassisModel` and `ChassisController` and has `ChassisController` implementations derive it from the supplied gearset.

### Benefits

Bug fix :)

### Possible Drawbacks

This is technically an API breaking change which I want to release as a patch version. I don't think it will break anyone's code, though, and the migration step would be to supply the old default max velocity manually.

### Verification Process

This was verified by hand.

### Applicable Issues

Closes #219.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/996bf665e68a1fb56c06a1dde902928931684306

PR #244, opened 2018-10-04
AsyncMotionProfileController does not account for external ratio

### LINKED ISSUE #243: AsyncMotionProfileController does not account for external ratio
`AsyncMotionProfileController` should factor in the external ratio from the `GearsetRatioPair`.

`leftRPM` and `rightRPM` in `AsyncMotionProfileController` are actually wheel RPM, not motor RPM. So we need to use the gear ratio to convert it.

Octogonapus:
### Description of the Change

`AsyncMotionProfileController` should factor in the external ratio from the `GearsetRatioPair`.

`leftRPM` and `rightRPM` in `AsyncMotionProfileController` are actually wheel RPM, not motor RPM. This PR uses the gear ratio to convert from wheel RPM to motor RPM. This PR also has `AsyncMotionProfileController` check if the gear ratio is got is zero, and throws a `std::invalid_argument` exception if that's the case.

### Benefits

More correct path following :)

### Possible Drawbacks

None.

### Verification Process

A test was added for the linear to rotational conversion.

### Applicable Issues

Closes #243.

### DIFF
--- a/include/okapi/api/control/async/asyncMotionProfileController.hpp
+++ b/include/okapi/api/control/async/asyncMotionProfileController.hpp
@@ -34,7 +34,8 @@ struct Point {
 class AsyncMotionProfileController : public AsyncPositionController<std::string, Point> {
   public:
   /**
-   * An Async Controller which generates and follows 2D motion profiles.
+   * An Async Controller which generates and follows 2D motion profiles. Throws a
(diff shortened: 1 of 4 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/7466d360da2beb1f118a16ba480dedf44c239e34

PR #245, opened 2018-10-05
Issue #234: Pass shared_ptr by const ref

### LINKED ISSUE #234: Pass shared_ptr by const ref
We can avoid an atomic increment-decrement by passing `std::shared_ptr` instances by const reference most of the time. This should be backward-compatible.

PixelDust22:
### Description of the Change

Used const refs to pass shared_ptr to functions

### Benefits

Avoid unnecessary reference count increment/decrement

### Possible Drawbacks

Didn't change factory classes

### Verification Process

- [x] Compiles
- [x] run cppcheck
- [x] run clang-format

### Applicable Issues

Closes #234

### DIFF
--- a/include/okapi/api/chassis/controller/chassisController.hpp
+++ b/include/okapi/api/chassis/controller/chassisController.hpp
@@ -26,7 +26,7 @@ class ChassisController : public ChassisModel {
    *
    * @param imodel underlying ChassisModel
    */
-  explicit ChassisController(std::shared_ptr<ChassisModel> imodel,
+  explicit ChassisController(const std::shared_ptr<ChassisModel> &imodel,
                              double imaxVelocity,
                              double imaxVoltage = 12000);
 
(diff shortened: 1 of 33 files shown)


Octogonapus (changes requested):
Some general feedback:
1. Say "Closes #foo" so GitHub will automatically close the related issue when the PR is merged.
2. Did you run clang-format? I don't think it would produce any changes here, but it's good to run it regardless.
3. Issues/PRs with the "needs docs" label require you to update the documentation before merging. The docs are on this branch: https://github.com/purduesigbots/pros-docs/tree/okapilib

I will live with not changing the factories yet. I should go make this change on #230.

PixelDust22:
Clang-format actually wrapped some more lines. Made another commit to reflect the changes.

Octogonapus (approved):

PR #247, opened 2018-10-06
Make Controller::operator[] return the same ControllerButton instance

### LINKED ISSUE #246: Controller::operator[] should not return a new ControllerButton instance
People tend to query a button once, so the natural API usage is something like `master[ControllerButton::X].changedToPressed()`. If you only query the button once, saving it to a variable is more verbose than inlining the button. Right now that isn't possible because `Controller::operator[]` return a new `ControllerButton` instance. If instead, we return the same instance, then both API usages are supported.

Octogonapus:
### Description of the Change

People tend to query a button once, so the natural API usage is something like `master[ControllerButton::X].changedToPressed()`. If you only query the button once, saving it to a variable is more verbose than inlining the button. Right now that isn't possible because `Controller::operator[]` return a new `ControllerButton` instance. If instead, we return the same instance, then both API usages are supported.

### Benefits

Described above :)

### Possible Drawbacks

Slightly more memory usage (12 static pointers) but it's not really a problem.

### Verification Process

This was tested by hand with
```cpp
  Controller master;
  std::cout << &master[ControllerDigital::X] << ", " << &master[ControllerDigital::X] << std::endl;
  std::cout << &master[ControllerDigital::A] << ", " << &master[ControllerDigital::A] << std::endl;
  std::cout << &master[ControllerDigital::L1] << ", " << &master[ControllerDigital::L1] << std::endl;
  pros::delay(100);
```
which prints
```
0x39fe220, 0x39fe220
0x39fe638, 0x39fe638
0x39fe648, 0x39fe648
```

### Applicable Issues

Closes #246.

### DIFF
--- a/include/okapi/impl/device/controller.hpp
+++ b/include/okapi/impl/device/controller.hpp
@@ -11,6 +11,7 @@
 #include "api.h"
 #include "okapi/impl/device/button/controllerButton.hpp"
 #include "okapi/impl/device/controllerUtil.hpp"
+#include <array>
 
 namespace okapi {
 class Controller {
@@ -108,8 +109,9 @@ class Controller {
   virtual std::int32_t getBatteryLevel();
 
   protected:
-  const ControllerId m_id;
+  ControllerId m_id;
   pros::Controller controller;
+  static std::array<ControllerButton *, 12> buttonArray;
 };
 } // namespace okapi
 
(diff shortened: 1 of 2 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/38c169711edf7e75775cb7d034de9e41613c836c

PR #251, opened 2018-10-16
Update kernel

Octogonapus:
### Description of the Change

This PR updates the kernel to v3.1.2.

- [x] Upgrade to v3.1.2 from the current prerelease version

### Benefits

v3.1.2 brings a bug fix for calling functions that use mutexes before they were initialized (in our case, `pros::Motor::get_position()`.

### Possible Drawbacks

Silent change on the Ultrasonic order, but I have swapped our internal order to be consistent.

### Verification Process

The mutex fix was verified by hand.

### Applicable Issues

None.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/cdd1b02aec6fc3681fe5cf61addae5885f936577

PR #257, opened 2018-10-30
Add a way to mirror ChassisController turns

### LINKED ISSUE #222: Easy mirroring for turns on ChassisController
I'm always frustrated when... I have to find my own solution to inverting my turn targets for making an auton routine work on both sides of the field. Previously, I always implemented this in my own turning functions, so it would be nice if there was a seamless way in okapilib.

Ideally, a good solution would be a method to reverse all turn targets on ChassisController or similar.

Octogonapus:
### Description of the Change

This PR adds the method `setTurnsMirrored()` so users can mirror their turns without creating a sign they manage themselves.

### Benefits

Cleaner user code.

### Possible Drawbacks

None.

### Verification Process

New tests were added.

### Applicable Issues

Closes #222.

### DIFF
--- a/include/okapi/api/chassis/controller/chassisController.hpp
+++ b/include/okapi/api/chassis/controller/chassisController.hpp
@@ -89,6 +89,13 @@ class ChassisController : public ChassisModel {
   virtual void turnAngleAsync(double idegTarget) = 0;
 
   /**
+   * Sets whether turns should be mirrored.
+   *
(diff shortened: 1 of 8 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/52da8efc681633e4a8e173e9425bba10bb8a75d1

PR #261, opened 2018-11-03
Add a method to APIC to set the zero point

### LINKED ISSUE #258: Add a method to APIC to set the zero point
User Willem Scholten requested that there be a method on APIC which lets you set the "zero point" for the controller. This is reasonable and would simply call `tarePosition()` on the underlying motor.

Octogonapus:
### Description of the Change

This PR adds a method `tarePosition()` which sets the "zero position" for the controller.

### Benefits

Exposes functionality to set the zero position.

### Possible Drawbacks

None.

### Verification Process

This change was not verified.

### Applicable Issues

Closes #258.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/c0123a15148508f53e0649202c977580d74e442f

PR #262, opened 2018-11-03
Add a method to APIC to interrupt a movement

### LINKED ISSUE #259: Add a stop method to APIC
Gabriel asked how to interrupt an APIC controller mid-movement and stop it. Disabling the controller will accomplish this, but then you'll need to manage the disabled lifecycle when making any movements. I think having a `stop()` method which just sets the velocity to zero will work fine.

Octogonapus:
### Description of the Change

This PR adds a method `stop()` to interrupt a motor mid-movement and stop it. Disabling the controller will accomplish this, but then you'll need to manage the disabled lifecycle when making any movements.

### Benefits

A nicer API to stop the controller.

### Possible Drawbacks

Stopping the controller and then disabling/enabling it will cause it to move back to its last set target. This is the intended behavior, but some users might get caught off-guard by it.

### Verification Process

This change was not verified.

### Applicable Issues

Closes #259.

### DIFF
--- a/include/okapi/api/control/async/asyncPosIntegratedController.hpp
+++ b/include/okapi/api/control/async/asyncPosIntegratedController.hpp
@@ -116,6 +116,11 @@ class AsyncPosIntegratedController : public AsyncPositionController<double, doub
    */
   virtual void setMaxVelocity(std::int32_t imaxVelocity);
 
+  /**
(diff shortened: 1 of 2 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/fd78f7c8979f8700e34f1052b6b7a5f96d1cf82d

PR #263, opened 2018-11-04
Reword reset docs

### LINKED ISSUE #260: Reword reset docs
A few users have been misled by the current reset docs, this should be reworded so that it clearly states that only the controller's internal state is reset.

Octogonapus:
### Description of the Change

A few users have been misled by the current reset docs. This PR rewords the docs so that it clearly states that only the controller's internal state is reset.

### Benefits

Clearer wording and less confused users.

### Possible Drawbacks

None.

### Verification Process

This change was not verified.

### Applicable Issues

Closes #260.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/9a564bd5331d3ec3b37eeb49db69496a4afa5971

PR #273, opened 2018-11-12
Add a method to remap gyro heading

### LINKED ISSUE #224: Add a way to get the gyro heading in +-1800
A few people have expressed interest in a way to get the gyro angle from `[-1800, 1800]`. This is a reasonable request we should support.

If the gyro heading is `[0, 3600]` then this should be implemented as:
```
(x - 1800) % 3600 - 1800
```
Which is equivalent to, but faster than:
```
remapRange((x - 1800) % 3600, 0, 3600, -1800, 1800)
```

Octogonapus:
### Description of the Change

This PR adds `getRemapped()` to get the gyro value with new bounds.

### Benefits

Users can get the gyro value in `+-1800` (or other bounds) easier.

### Possible Drawbacks

None.

### Verification Process

This was not verified.

### Applicable Issues

Closes #224.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/9865cf61dfe52268f72865a0a4b33c8dbb51cf8b

PR #287, opened 2018-11-25
Rename imev5TPR

### LINKED ISSUE #286: Rename imev5TPR
`imev5TPR` is poorly named because it only makes sense for one gearset. We should have a TPR for all three gearsets.

Octogonapus:
### Description of the Change

This PR renames the poorly named `imev5TPR` so it reflects the gearset it corresponds to.

### Benefits

Better constant names.

### Possible Drawbacks

None.

### Verification Process

This was not verified.

### Applicable Issues

Closes #286.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/120ecb5f4cf9c8e7ec7e4d185d73a5eb0c5ab5e4

PR #292, opened 2018-12-10
Add two more bounds for the controllerSet target

### LINKED ISSUE #291: IVPIDC computes controllerSet target incorrectly
`IterativeVelPIDController` computes the `controllerSet()` target as `target = remapRange(ivalue, -1, 1, outputMin, outputMax);` but this should remap `[-1, 1]` into a max and min for the target so it can be scaled into a range more like `[-200, 200]`.

Octogonapus:
### Description of the Change

This PR adds two more target bounds so that `IterativeVelPIDController::controllerSet()` can scale its target independently of the output limits.

### Benefits

Better target scaling.

### Possible Drawbacks

None.

### Verification Process

Added a test.

### Applicable Issues

Closes #291.

### DIFF
--- a/include/okapi/api/control/async/asyncWrapper.hpp
+++ b/include/okapi/api/control/async/asyncWrapper.hpp
@@ -141,6 +141,17 @@ class AsyncWrapper : virtual public AsyncController<Input, Output> {
   }
 
   /**
+   * Sets the (soft) limits for the target range that controllerSet() scales into. The target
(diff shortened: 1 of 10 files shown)


Octogonapus:
Docs: ~~https://github.com/purduesigbots/pros-docs/commit/187467d7d07dcabe8d70ccc07f8db496bf36c112~~
https://github.com/purduesigbots/pros-docs/commit/e01eeeff1e55b4d8f1966b9bbc033931d47ae43e

PR #323, opened 2019-01-20
Derive ControllerInput<bool> for AbstractButton

lePerdu:
### Description of the Change

Buttons have a trivial implementation of `ControllerInput<bool>`: `isPressed()`. Adding this implementation allows using buttons (and limit switches, etc.) as inputs to motor controllers in user code.

### Benefits

This allows users of the library to create motor controllers which take boolean sensors as input.

### Possible Drawbacks

N/A

### Verification Process

The code is simple enough that it can be verified visually. This doesn't remove or change functionality, so it cannot introduce any regressions.

### Applicable Issues

N/A

### DIFF
--- a/include/okapi/api/device/button/abstractButton.hpp
+++ b/include/okapi/api/device/button/abstractButton.hpp
@@ -7,8 +7,10 @@
  */
 #pragma once
 
+#include "okapi/api/control/controllerInput.hpp"
+
 namespace okapi {
-class AbstractButton {
+class AbstractButton : public ControllerInput<bool> {
   public:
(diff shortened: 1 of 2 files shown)


Octogonapus (changes requested):
Looks good. Please target this at `feature/v3.3.9` and make a corresponding docs PR against `https://github.com/purduesigbots/pros-docs/tree/okapilib`.

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/pull/91

Octogonapus (approved):

PR #326, opened 2019-02-03
Add lazy logging

### LINKED ISSUE #324: Guard logger calls
All logger calls should be guarded for performance reasons. Add methods to `Logger` like `isInfoLevelEnabled()` for use in guarding log statements.

Octogonapus:
### Description of the Change

This PR adds lazy logging using lambdas. Log messages are now inside lambdas so they can be evaluated only if the log level is turned on.

### Benefits

Log messages are no longer constructed regardless of whether that log level is turned on, so logging can be used more frequently.

### Possible Drawbacks

None.

### Verification Process

I sprinkled in some log statements to some integration tests, everything appears to work.

### Applicable Issues

Closes #324.

### DIFF
--- a/include/okapi/api/control/async/asyncWrapper.hpp
+++ b/include/okapi/api/control/async/asyncWrapper.hpp
@@ -70,7 +70,7 @@ class AsyncWrapper : virtual public AsyncController<Input, Output> {
    * Sets the target for the controller.
    */
   void setTarget(const Input itarget) override {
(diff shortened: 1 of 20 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/e1c9e5b2ea9fb83b860dee567522764e004f0ef9

PR #327, opened 2019-02-04
Fix CCPID settling without stopping motors

### LINKED ISSUE #93: Create better MockMotor
The `MockMotor` is currently just used to check for controller housekeeping. This is useful, but there is room for improvement in the mock. I think we can properly simulate the motor in another thread and get somewhat accurate behavior.

### LINKED ISSUE #308: CCPID can settle without stopping motors
CCPID can exit from `waitUntilSettled()` without stopping its motors because of a scheduling conflict between the thread running `waitUntilSettled()` and the internal thread.

Octogonapus:
### Description of the Change

This PR fixes `ChassisControllerPID` settling without stopping the motors. Sometimes, a race condition between the task waiting to settle and the internal CCPID task caused the motors to get powered just after settling (i.e. the internal task did one more loop after settling).

### Benefits

No more drifting robots.

### Possible Drawbacks

None.

### Verification Process

I could only reproduce this issue in a small capacity on my test bed. Others in the community have tested this with actual robots and report the problem as fixed with no regressions.

### Applicable Issues

Closes #308. Closes #93 (mostly).


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/43de51061a68424d3062b30f35ae54e20791b38b

PR #334, opened 2019-02-10
Fix CCPID using the same Rate in multiple places

### LINKED ISSUE #332: IterativePosPIDController doesn't handle the kD gain properly.
**Describe the bug**
An IterativePosPIDController will speed up as it gets closer to the target according to the kD value, not slow down as it should. This is because the IterativePosPIDController substracts the derivative term instead of adding it. The difference in error will already be negative, so this will negate a negative and make the robot move faster as it approaches the target.

**Steps To Reproduce**
1. Make a ChassisControllerPID with distance gains of [0.002, 0, 0.008].
2. The robot will not slow down as it approaches the target
3. Negate the D gain
4. The robot will slow down as it approaches the target

**Expected behavior**
I expected the derivative term to be added, slowing down the robot as it the error gets closer to 0.

**Versions**
I am on OkapiLib 3.3.9 on PROS version 3.1.5.

**Additional Information**
No additional information.

Octogonapus:
### Description of the Change

This PR fixes a bug which caused the IPPC derivative term to be zero most of the time, causing people to think the derivative term did not work. This symptom is a red herring. The real problem was that CCPID re-used `rate` in multiple places, causing some calls to `delayUntil` to not delay. This PR also fixes CCPID starving the LVGL task.

### Benefits

^ :)

### Possible Drawbacks

`getError()` no longer updates while disabled. This has been specified in the docs.

### Verification Process

This was verified by hand with a real robot.

### Applicable Issues

Closes #332.

### DIFF
--- a/include/okapi/api/chassis/controller/chassisControllerPid.hpp
+++ b/include/okapi/api/chassis/controller/chassisControllerPid.hpp
@@ -127,7 +127,6 @@ class ChassisControllerPID : public virtual ChassisController {
   protected:
   std::shared_ptr<Logger> logger;
   TimeUtil timeUtil;
-  std::unique_ptr<AbstractRate> rate;
   std::unique_ptr<IterativePosPIDController> distancePid;
   std::unique_ptr<IterativePosPIDController> turnPid;
   std::unique_ptr<IterativePosPIDController> anglePid;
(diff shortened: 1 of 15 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/c5275e308943f7e8756ff0254f3bd9d7f6fefc39

PR #342, opened 2019-02-12
Add CCPID voltage mode

### LINKED ISSUE #338: Add an option to run CCPID in voltage mode
Add a flag in CCPID (and expose it in the builder) to run the control loop in either velocity mode (which should be the default, as it works now) or voltage mode.

Octogonapus:
### Description of the Change

This PR adds a flag to CCPID to run the control loop in voltage mode, instead of in velocity mode (which is the default now).

### Benefits

Some users would rather not run the PID controller outputs through the built-in velocity control.

### Possible Drawbacks

None.

### Verification Process

`driveVectorVoltage(0, x)` has been tested to be equivalent to `rotate(x)`.

### Applicable Issues

Closes #338.

### DIFF
--- a/include/okapi/api/chassis/controller/chassisController.hpp
+++ b/include/okapi/api/chassis/controller/chassisController.hpp
@@ -104,7 +104,7 @@ class ChassisController : public ChassisModel {
    *
    * @param ipower motor power
    */
-  void forward(double ispeed) const override;
+  void forward(double ispeed) override;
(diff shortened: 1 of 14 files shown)


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/b957c8e50bf713e62add0986b55fa3cfe90569a1

PR #349, opened 2019-02-16
Use new task notify changes to handle killing internal tasks

### LINKED ISSUE #249: CCPID can segfault if its task is destroyed
If a task is deleted (like auton or opcontrol when the robot is disabled) then the idle task will eventually free that task's stack. Any tasks that require memory from that stack (like CCPID's internal thread which requres an implict `this`) will then cause a segfault.

### LINKED ISSUE #321: Motion Profile Memory Permission Error
**Describe the bug**
In the middle of the autonomous period, sometimes when I switched the competition switch to disable, the Brain's screen will display an error box saying "Memory Permission Error"  

**Steps To Reproduce**
Create a motion profile controller, generate a path, and run the path; then disable the bot in the middle of the period. 

**Expected behavior**
The screen should not display the memory permission error. 

**Versions**
kernel    3.1.5      pros-mainline
okapilib  3.3.8      pros-mainline

**Additional Information**
```c++
profileController.generatePath({Point{0_ft, 0_ft, 0_deg}, Point{3.8_ft, 0_ft, 0_deg}}, "A");
profileController.setTarget("A");
profileController.waitUntilSettled();
```

Octogonapus:
### Description of the Change

This PR uses the new kernel's `task_notify_when_deleting()` changes to kill internal tasks when their parent task dies. This does not directly fix the segfault problem when an internal task references a parent object on another task's stack but does fix "runaway" tasks, such as a CCPID in `autonomous()` which keeps running after auton ends. The real fix for the segfaults are the builders, which force users to allocate their objects on the heap. This way, the parent objects will never be freed out from under the internal tasks, so no segfaults can happen (it does cause runaway tasks, though, which this PR fixes).

### Benefits

No more runaway tasks.

### Possible Drawbacks

None.

### Verification Process

This was verified by hand using the "Timed Run" mode to simulate a match. This was also verified by hand using a variety of task setups described in https://github.com/purduesigbots/pros/pull/100.

### Applicable Issues

Closes #249. Closes #321.


Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/2ee360f47d8fa78ee63496e4f983b7791bee0091

PR #355, opened 2019-03-12
Initialize member variables in VelMath

theol0403:
### Description of the Change
Initialize default values for `VelMath` so calling `step()` with dT of 0 will not result in undefined behavior.

### Benefits
Prevent `VelMath` from reporting erroneous velocity values until minimum dT is satisfied.

### Possible Drawbacks
None.

### Verification Process
Add a test to confirm `getVelocity()` and `getAccel()` return 0 after initialization.

### DIFF
--- a/include/okapi/api/filter/velMath.hpp
+++ b/include/okapi/api/filter/velMath.hpp
@@ -78,7 +78,7 @@ class VelMath {
   Logger *logger;
   QAngularSpeed vel{0_rpm};
   QAngularSpeed lastVel{0_rpm};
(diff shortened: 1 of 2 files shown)


theol0403:
I made the mistake of writing `{0_rpm / 0_s}` which results in `nan`.
Fixes are one of three:
1. `{0_rpm / 1_s}`
2. `{0_rpm / second}`
3. `{0.0}`

I have implemented the third option. The initially planned `{0}` gives an "ambiguous constructor" error, but `{0.0}` works.

Octogonapus (approved):

Issue #358, opened 2019-04-16
Allow saving and loading of Pathfinder paths to files


juliaschatz:
**Is your feature request related to a problem? Please describe.**
Pathfinder paths cannot be stored in files, which limits the number of paths available.

**Describe the solution you'd like**
Implement Pathfinder I/O to save and load paths from files.

**Describe alternatives you've considered**
Generating paths on the fly- Too slow
Generating paths on initialize- Limited by memory

juliaschatz:
I'm working on implementing this- should be pretty simple as this functionality is in the original Pathfinder.

PR #359, opened 2019-04-16
Issue #358: Implement saving/loading paths

juliaschatz:
Create methods in asyncMotionProfileController.
Needs testing, error checking and sanitization should be implemented.

### Description of the Change
Implements Pathfinder serialize/deserialize methods in asyncMotionProfileController

### Motivation
Paths are memory-heavy, and the v5 brain is limited. This would allow many more paths to be used as they would not have to be generated on the fly or a more limited number generated on initialize.

### Possible Drawbacks
Completely optional feature, no drawbacks foreseen.

### Verification Process
Testing WIP

### DIFF
--- a/src/api/control/async/asyncMotionProfileController.cpp
+++ b/src/api/control/async/asyncMotionProfileController.cpp
@@ -8,6 +8,7 @@
 #include "okapi/api/control/async/asyncMotionProfileController.hpp"
 #include "okapi/api/util/mathUtil.hpp"
 #include <numeric>
+#include <algorithm>
 
 namespace okapi {
 AsyncMotionProfileController::AsyncMotionProfileController(
@@ -361,35 +362,53 @@ CrossplatformThread *AsyncMotionProfileController::getThread() const {
 }
 
 void AsyncMotionProfileController::storePath(std::string idirectory, std::string ipathId) {
-  // todo this makes the assumption that idirectory doesn't have a trailing / and that
-  // ipathId is a valid part of a filename-which is not necessarily true
-  FILE* leftPathFile = fopen((idirectory + "/" + ipathId + ".left.csv").c_str(), "w");
-  FILE* rightPathFile = fopen((idirectory + "/" + ipathId + ".right.csv").c_str(), "w");
-  // todo file error checking
+  FILE* leftPathFile = fopen(makePath(idirectory, ipathId + ".left.csv").c_str(), "w");
+  FILE* rightPathFile = fopen(makePath(idirectory, ipathId + ".right.csv").c_str(), "w");
 
-  // todo make sure this is an actual loaded path that exists
+  // Make sure we can open the file successfully
+  if (leftPathFile == NULL || rightPathFile == NULL) {
+    logger->warn(
+      "AsyncMotionProfileController: Couldn't store path " +
+      ipathId + " in directory " + idirectory);
+    return;
+  }
+
   auto pathData = this->paths.find(ipathId);
-  int len = pathData->second.length;
 
-  // Serialize paths
-  pathfinder_serialize_csv(leftPathFile, pathData->second.left, len);
-  pathfinder_serialize_csv(rightPathFile, pathData->second.right, len);
+  // Make sure path exists
+  if (pathData == paths.end()) {
+    logger->warn(
+      "AsyncMotionProfileController: Controller was asked to serialize non-existent path " +
+      ipathId);
+    // Do nothing- can't serialize nonexistent path
+  }
+  else {
+    int len = pathData->second.length;
+
+    // Serialize paths
+    pathfinder_serialize_csv(leftPathFile, pathData->second.left, len);
+    pathfinder_serialize_csv(rightPathFile, pathData->second.right, len);
+  }
   
   fclose(leftPathFile);
   fclose(rightPathFile);
 }
 
 void AsyncMotionProfileController::loadPath(std::string idirectory, std::string ipathId) {
-  // todo this makes the assumption that idirectory doesn't have a trailing / and that
-  // ipathId is a valid part of a filename-which is not necessarily true
-  FILE* leftPathFile = fopen((idirectory + "/" + ipathId + ".left.csv").c_str(), "w");
-  FILE* rightPathFile = fopen((idirectory + "/" + ipathId + ".right.csv").c_str(), "w");
+  FILE* leftPathFile = fopen(makePath(idirectory, ipathId + ".left.csv").c_str(), "r");
+  FILE* rightPathFile = fopen(makePath(idirectory, ipathId + ".right.csv").c_str(), "r");
 
-  // todo file error checking
+  // Make sure we can open the file successfully
+  if (leftPathFile == NULL || rightPathFile == NULL) {
+    logger->warn(
+      "AsyncMotionProfileController: Couldn't load path " +
+      ipathId + " in directory " + idirectory + ". Make sure the file exists.");
+    return;
+  }
 
   // Count lines in file, remove one for headers
   int count = 0;
-  for (char c = getc(leftPathFile); c != EOF; c = getc(leftPathFile)) {
+  for (int c = getc(leftPathFile); c != EOF; c = getc(leftPathFile)) {
     if (c == '\n') {
       ++count;
     }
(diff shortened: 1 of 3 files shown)


juliaschatz:
Still todo: File IO toddler-proofing, path error checking.

HotelCalifornia:
note: [heap size was recently increased in the PROS kernel](https://github.com/purduesigbots/pros/pull/89/files#diff-3e67e7437ae7de6cc9138170266edcda), so if paths are being defined as stack variables in a task one easy way to try circumventing memory issues would be to bump the stack size used for said task.

That said:
1. I don't know exactly how memory-intensive pathfinder paths are, so even with the above increase and workaround it's possible this would be needed
2. The workaround mentioned above would not work for the competition tasks (initialize, competition_initialize, etc) as their stacks are statically defined 
3. This seems like it would be a good thing to have anyway, regardless of my above comments that may seem to the contrary

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -328,4 +328,50 @@ void AsyncMotionProfileController::startThread() {
...
+    if (c == '\n') {
+      ++count;
+    }
+  }
+  count--;
+  rewind(leftPathFile);
+
+  // Allocate memory
+  auto *leftTrajectory = (Segment *)malloc(sizeof(Segment) * count);
+  auto *rightTrajectory = (Segment *)malloc(sizeof(Segment) * count);
```
HotelCalifornia, line 451: is there a special reason these aren't allocated using `operator new[]`?

Octogonapus: They are doing it the same way it's already implemented in my code. And I did it that way because it's how Jaci did it in her tutorial.

Octogonapus:
@nickschatz I am inclined to put this on v4.0.0, which means it would be released after worlds. Even if I put it on v3.3.13, it would probably still come out after worlds, though, so do you mind if it's on v4.0.0?

juliaschatz:
> @nickschatz I am inclined to put this on v4.0.0, which means it would be released after worlds. Even if I put it on v3.3.13, it would probably still come out after worlds, though, so do you mind if it's on v4.0.0?

Anytime after worlds is the same to me. If I end up needing this I'll just build okapilib for myself.

@HotelCalifornia if I'm reading that right, then the memory available to the user program is 48 MB? Since I can't seem to find it in PROS, what's the default stack size?

My ballpark for a 7 second path is about 44000 bytes, assuming 64 bit doubles. 
If that's all correct, I don't think this change is essential, but would still be good at some point. My impetus was some rumbling about deleting paths because of running out of memory, but I don't have any concrete reports.

baylessj:
I'm definitely in favor of a change that allows for saving more Pathfinder paths at any given time; both of the BLRS robots use Pathfinder as the basis for all of their movements and as a result I've struggled to generate/delete paths at the right times to make that possible. 

This sounds to me like the most robust solution for increasing available memory for paths, and I would recommend using these functions to the BLRS team in future seasons. One simpler approach for reducing memory use that I found was reducing the time step for the paths from 1ms to 10ms; I noticed no change in performance from running the path at the motor's update rate as opposed to the faster 1ms rate. Until this change gets fully implemented and tested the time step change may serve as an easier improvement.

juliaschatz:
@baylessj Wow, definitely. The path I was referencing (generated by a friend) had a 20 ms timestep, which would put the byte estimate for an Okapi path of the same length at a staggering 860 kb. I agree there's absolutely no reason to go faster than 10 ms.

juliaschatz:
I believe this is feature complete now. I'll try to test it tonight, let me know if anything should be changed or added.

Octogonapus:
~~I'm putting this on v4.0.0 then. Would one of @baylessj or @nickschatz please open an issue to change the dT to 10ms?~~ And @nickschatz please rebase off of `feature/v4.0.0`.

Edit: Opened in #362.

src/api/control/async/asyncMotionProfileController.cpp
HotelCalifornia: figured that was the case, just wanted to make sure

HotelCalifornia:
> if I'm reading that right, then the memory available to the user program is 48 MB? Since I can't seem to find it in PROS, what's the default stack size?

[see docs](https://pros.cs.purdue.edu/v5/api/c/rtos.html#task-stack-depth-default)

Octogonapus:
@nickschatz You are getting a weird conflict so I merged `feature/v4.0.0` up with `develop`. Please pull those changes into your branch.

juliaschatz:
Got my SD cards in so I did some testing on the v5. Everything seems to work well. There's not an easy way to unit test the actual saving/loading of files, but I can probably hack something together if you think it's necessary.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -359,4 +361,116 @@ void AsyncMotionProfileController::startThread() {
...
+  FILE *rightPathFile = fopen(rightFilePath.c_str(), "w");
+
+  // Make sure we can open the file successfully
+  if (leftPathFile == NULL) {
+    LOG_WARN("AsyncMotionProfileController: Couldn't open file " + leftFilePath + " for writing");
+    return;
+  }
+  if (rightPathFile == NULL) {
+    LOG_WARN("AsyncMotionProfileController: Couldn't open file " + rightFilePath + " for writing");
+    return; // cppcheck thinks this is a resource leak, but fclose(NULL) is undefined behavior
```
Octogonapus: And cppcheck is right because `leftPathFile` is not null.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -359,4 +361,116 @@ void AsyncMotionProfileController::startThread() {
...
+  FILE *rightPathFile = fopen(rightFilePath.c_str(), "r");
+
+  // Make sure we can open the file successfully
+  if (leftPathFile == NULL) {
+    LOG_WARN("AsyncMotionProfileController: Couldn't open file " + leftFilePath + " for reading");
+    return;
+  }
+  if (rightPathFile == NULL) {
+    LOG_WARN("AsyncMotionProfileController: Couldn't open file " + rightFilePath + " for reading");
+    return; // cppcheck thinks this is a resource leak, but fclose(NULL) is undefined behavior
```
Octogonapus: And `cppcheck` is right because leftPathFile is not null.

Octogonapus (changes requested):
I think this won't be so hard to test. The `Logger` tests open an in-memory file, maybe you can do the same here:
1. Make a path as the other tests do
2. Serialize it to an in-memory file
3. Deserialize it
4. Run it and assert some things as the other tests do

Edit: Also make sure you have run clang-format

src/api/control/async/asyncMotionProfileController.cpp
juliaschatz: Good catch, fixed

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -359,4 +361,140 @@ void AsyncMotionProfileController::startThread() {
...
+  removePath(ipathId);
+
+  paths.emplace(ipathId, TrajectoryPair{leftTrajectory, rightTrajectory, count});
+}
+
+std::string AsyncMotionProfileController::makeFilePath(std::string directory,
+                                                       std::string filename) {
+  std::string path(directory);
+
+  // Make sure this is an absolute path beginning /usd/<subdirs>
```
Octogonapus: Stray comment after a refactor?

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -21,12 +21,31 @@ class MockAsyncMotionProfileController : public AsyncMotionProfileController {
     AsyncMotionProfileController::executeSinglePath(path, std::move(rate));
   }
 
+  static std::string makeFilePath(std::string directory, std::string filename) {
```
Octogonapus: What is the purpose of these redefinitions? If you just want to make them public you can `using` them like above.

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -258,3 +288,49 @@ TEST_F(AsyncMotionProfileControllerTest, FollowPathMirrored) {
...
+
+  controller->removePath("A");
+  controller->internalLoadPath(leftPathFile, rightPathFile, "A");
+  EXPECT_EQ(controller->getPaths().front(), "A");
+  EXPECT_EQ(controller->getPaths().size(), 1);
+  EXPECT_EQ(controller->getPathData("A").length, genPathLen);
+
+  controller->setTarget("A");
+  EXPECT_EQ(controller->getTarget(), "A");
+}
```
Octogonapus: Please end files with a newline.

Octogonapus (changes requested):

Octogonapus (changes requested):
This looks good. Now you just need to do the docs side for the two public methods you added against this: https://github.com/purduesigbots/pros-docs/tree/okapilib-4.0.0

juliaschatz:
Done at https://github.com/nickschatz/pros-docs/tree/pf-save-load-doc but I'm unable to get pros-docs to build. Looks like an internal sphinx error that I can mess around with sometime soon.

Octogonapus:
Docs: https://github.com/purduesigbots/pros-docs/commit/6eb00be1cefb08167ac625a805f071dd57560f6b

Octogonapus (approved):

PR #363, opened 2019-04-29
Change motion profile DT to 10 milliseconds

### LINKED ISSUE #362: Change motion profile dt to 10ms
AMPC and ALMPC generate paths with a dt of 1ms. There is no performance difference between 1ms and 10ms, so we should change it to 10ms to reduce path complexity and memory needed.

juliaschatz:
### Description of the Change
Changes the default motion profile delta-time to 10 milliseconds, and changes the follower to read the delta-time from the path instead of assuming it's the default.

### Motivation
10 ms is the fastest the motors can update, it bogs down memory and path generation speed to do a shorter delta-time.

### Possible Drawbacks
This code is 100% backwards-compatible. No reduced performance in path following should exist.

### Verification Process
Tests pass. I also made this change before Worlds and ran it (the AMPC change) on one of our robots.

### Applicable Issues

Closes #362

### DIFF
--- a/src/api/control/async/asyncLinearMotionProfileController.cpp
+++ b/src/api/control/async/asyncLinearMotionProfileController.cpp
@@ -213,13 +213,13 @@ void AsyncLinearMotionProfileController::executeSinglePath(const TrajectoryPair 
   const auto reversed = direction.load(std::memory_order_acquire);
 
   for (int i = 0; i < path.length && !isDisabled(); ++i) {
-    float segDT = path.segment[i].dt;
+    const auto segDT = path.segment[i].dt * second;
     currentProfilePosition = path.segment[i].position;
 
     const auto motorRPM = convertLinearToRotational(path.segment[i].velocity * mps).convert(rpm);
     output->controllerSet(motorRPM / toUnderlyingType(pair.internalGearset) * reversed);
 
-    rate->delayUntil(segDT * second);
+    rate->delayUntil(segDT);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/api/control/async/asyncLinearMotionProfileController.cpp
```diff
@@ -213,12 +213,13 @@ void AsyncLinearMotionProfileController::executeSinglePath(const TrajectoryPair
   const auto reversed = direction.load(std::memory_order_acquire);
 
   for (int i = 0; i < path.length && !isDisabled(); ++i) {
+    float segDT = path.segment[i].dt;
```
Octogonapus: I probably should have done this to begin with...

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -253,23 +253,25 @@ void AsyncMotionProfileController::executeSinglePath(const TrajectoryPair &path,
 
   if (followMirrored) {
     for (int i = 0; i < path.length && !isDisabled(); ++i) {
+      float segDT = path.left[i].dt;
```
Octogonapus: So I guess we assume that the left and right sides have the same dt? Seems reasonable to me but I could be persuaded against this.

Octogonapus (approved):

src/api/control/async/asyncLinearMotionProfileController.cpp
Octogonapus: After looking at this again, I think it would be better to have
```suggestion
    const auto segDT = path.segment[i].dt * second;
```

The reason being that this method leaves no room for error in interpreting the units of `path.segment[i].dt`. Same change for below.

Octogonapus (changes requested):

src/api/control/async/asyncMotionProfileController.cpp
juliaschatz: I mean, unless you want to run a different task for each side. I think it's a totally reasonable assumption to make- Pathfinder only generates paths that have a constant DT that's the same for both sides. It would be a massive rewrite that's far more complex for no benefit.

Octogonapus (approved):

PR #366, opened 2019-05-15
Prevent segfault when running path

### LINKED ISSUE #360: AMPC's removePath function can segfault if path to free is running
**Describe the bug**
Calling `AsyncMotionProfileController::removePath()` to free the currently running path can cause a segmentation fault.

**Steps To Reproduce**
Set a target to the AMPC and then free the target before the path has completed.

**Expected behavior**
The controller should be disabled prior to the path's removal ideally, or the removePath function could refuse to remove the path and warn the user about the conflict.

**Versions**
PROS: 3.1.6
OkapiLib: 3.3.8

**Additional Information**
The reason I discovered this issue was because I subclassed AMPC and added a custom waitUntilSettled() implementation that allowed for exiting the movement prior to the end of the path, in case a line was detected or the vision sensor returned something useful or whatnot. In order to save space for paths I free the path soon after the path completes, and in the case of an early exit I would get a segfault when freeing the path.

juliaschatz:
### Description of the Change
Fixes an error where a path being removed while being ran causes a segfault by preventing the path from being removed when it's running

### Motivation
Segfaults are bad

### Possible Drawbacks
This changes the behavior of removePath in a non-obvious way.

### Verification Process

Additional unit tests

### Applicable Issues

Fixes #360


juliaschatz:
Need to figure out what the behavior should be in Okapilib if the path can't be removed.

include/okapi/api/control/async/asyncMotionProfileController.hpp
```diff
@@ -79,10 +79,13 @@ class AsyncMotionProfileController : public AsyncPositionController<std::string,
 
   /**
    * Removes a path and frees the memory it used.
+   * This function returns true if the path was either deleted or didn't exist in the first place.
+   * It returns false if the path could not be removed because it is running.
    *
    * @param ipathId A unique identifier for the path, previously passed to generatePath()
+   * @return Whether the path no longer exists
```
Octogonapus:
```suggestion
   * @return True if the path no longer exists
```

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -175,13 +175,19 @@ std::string AsyncMotionProfileController::getPathErrorMessage(const std::vector<
                          [&](std::string a, Waypoint b) { return a + ", " + pointToString(b); });
 }
 
-void AsyncMotionProfileController::removePath(const std::string &ipathId) {
+bool AsyncMotionProfileController::removePath(const std::string &ipathId) {
+  if (isRunning.load(std::memory_order_acquire) && getTarget() == ipathId) {
+    LOG_WARN("Attempted to remove currently running path " + ipathId);
+    return false;
+  } 
   auto oldPath = paths.find(ipathId);
```
Octogonapus, line 186: Please put a newline after the `}`.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -175,13 +175,19 @@ std::string AsyncMotionProfileController::getPathErrorMessage(const std::vector<
...
   if (oldPath != paths.end()) {
     free(oldPath->second.left);
     free(oldPath->second.right);
     paths.erase(ipathId);
   }
+  // Return true whether we actually did anything or not
```
Octogonapus: Please put a newline after the `}`.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -251,29 +257,24 @@ void AsyncMotionProfileController::executeSinglePath(const TrajectoryPair &path,
...
-      model->right(leftRPM / toUnderlyingType(pair.internalGearset) * reversed);
-
-      rate->delayUntil(segDT);
+  for (int i = 0; i < path.length && !isDisabled(); ++i) {
+    const auto segDT = path.left[i].dt * second;
+    const auto leftRPM = convertLinearToRotational(path.left[i].velocity * mps).convert(rpm);
+    const auto rightRPM = convertLinearToRotational(path.right[i].velocity * mps).convert(rpm);
+
+    double rightSpeed = rightRPM / toUnderlyingType(pair.internalGearset) * reversed;
+    double leftSpeed = leftRPM / toUnderlyingType(pair.internalGearset) * reversed;
```
Octogonapus: Please make `rightSpeed` and `leftSpeed` `const`.

Octogonapus (changes requested):
Returning a flag if the remove failed seems like a fine solution to me.

juliaschatz:
Any thought about changes to internal uses of removePath? Should everything that expects a path to be removed error out if it isn't removed?

Octogonapus:
> Any thought about changes to internal uses of removePath? Should everything that expects a path to be removed error out if it isn't removed?

@nickschatz I think the changes to `removePath` should be fine (it shouldn't always return `true`, though). It should be a best-effort remove. We will have to test that its safety mechanism isn't falsely triggered when its called by another internal method.

Octogonapus:
@nickschatz Are you still working on this?

juliaschatz:
I kinda forgot about this when the semester ended. Give me a day or two and I'll have something.

juliaschatz:
The most key internal use of removePath is in generatePath (and loadPath) where it could be called by an end-user while running another path. For the other one (moveTo) I'll just probably ignore the result, but I have three thoughts about what should happen if a path is generated or loaded while a path of the same name is being ran.

The options are:
- Throw a runtime error and drop out of the program. This is the safest option and produces no ill side effects except possibly killing your entire robot in the autonomous routine.
- Throw a warning and do nothing. The currently running path will not be interrupted or replaced. If the path is ran again, it will be the old path and not the newly generated or loaded path. So it will be the incorrect behavior, but won't kill your robot.
- Throw a warning, kill the current movement, and replace it with the new path (without running the new path).

juliaschatz:
I chose the first option. Still need to write tests and documentation.

Octogonapus:
> Throw a runtime error and drop out of the program. This is the safest option and produces no ill side effects except possibly killing your entire robot in the autonomous routine.

No. OkapiLib promises to never kill your robot, except in one-time initialization things like constructors where you are basically guaranteed to catch it in testing.

For the other two options we need to think about why this issue came up.

> Throw a warning and do nothing. The currently running path will not be interrupted or replaced. If the path is ran again, it will be the old path and not the newly generated or loaded path. So it will be the incorrect behavior, but won't kill your robot.

This seems "just ok" but it depends on why this issue came up in the first place.

> Throw a warning, kill the current movement, and replace it with the new path (without running the new path).

This seems like the best option to me. I think this issue came up because people were removing paths before running new ones to save memory space (at least BLRS was).

include/okapi/api/control/async/asyncLinearMotionProfileController.hpp
```diff
@@ -267,5 +267,12 @@ class AsyncLinearMotionProfileController : public AsyncPositionController<std::s
 
   std::string
   getPathErrorMessage(const std::vector<Waypoint> &points, const std::string &ipathId, int length);
+
+  /**
+   * Attempts to remove a path without stopping execution, then if that fails, stops exection and removes the path
+   * Returns true if the path was successfully removed
+   * todo expose this?
```
Octogonapus: Nit: Make `todo` `TODO: ` because then CLion knows it's a todo and will highlight it in yellow and make them searchable

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -181,7 +176,7 @@ std::string AsyncMotionProfileController::getPathErrorMessage(const std::vector<
 }
 
 bool AsyncMotionProfileController::removePath(const std::string &ipathId) {
-  if (isRunning.load(std::memory_order_acquire) && getTarget() == ipathId) {
+  if (!isDisabled() && isRunning.load(std::memory_order_acquire) && getTarget() == ipathId) {
```
Octogonapus, line 179: Why force the controller to be enabled? I think removing a path from a disabled controller is fine.

juliaschatz: This if branch should only activate if a) The controller is enabled (`!isDisabled()`), b) A path is currently running and c) if the path being removed is the currently running path. So it is able to remove any path it wants from the disabled controller.

juliaschatz:
@Octogonapus what do you think about exposing `safeRemovePath`? I intended it as a utility method but see no reason not to expose it alongside (or maybe instead of) `removePath`, unless you want to enforce users managing their own path life cycles.

src/api/control/async/asyncMotionProfileController.cpp
Octogonapus: I see, I didn't look at the entire method body.

Octogonapus:
> what do you think about exposing safeRemovePath?

I think it is fine to expose it as well. I'd call it `forceRemovePath`, though, because it's not really "safe" (it's guaranteed to delete the path whether its being followed or not).

Additionally, have we considered the race conditions here? If we disable the controller and remove a path, is there any situation in which the code following the path could try to reference the path one last loop? That would be a problem.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -175,13 +175,22 @@ std::string AsyncMotionProfileController::getPathErrorMessage(const std::vector<
...
+  }
+
   auto oldPath = paths.find(ipathId);
   if (oldPath != paths.end()) {
     free(oldPath->second.left);
     free(oldPath->second.right);
     paths.erase(ipathId);
   }
+
+  // Return true whether the path was actually removed or not
```
Octogonapus: These comments are wrong now because we return false above.

include/okapi/api/coreProsAPI.hpp
```diff
@@ -94,3 +98,29 @@ class CrossplatformThread {
 
   CROSSPLATFORM_THREAD_T thread;
 };
+
+class CrossplatformMutex {
```
Octogonapus, line 102: Yep, this is what I had in mind. The PROS support for mutex should come around the fall semester.

juliaschatz:
https://github.com/purduesigbots/pros-docs/pull/134 docs pr

include/okapi/api/control/async/asyncLinearMotionProfileController.hpp
```diff
@@ -224,6 +227,14 @@ class AsyncLinearMotionProfileController : public AsyncPositionController<std::s
    */
   CrossplatformThread *getThread() const;
 
+  /**
+   * Attempts to remove a path without stopping execution, then if that fails,
+   * disables the controller and removes the path
+   *
+   * @param ipathId The path ID that will be removed
+   */
+  void forceRemovePath(std::string ipathId);
```
Octogonapus: Should be a const ref

include/okapi/api/control/async/asyncMotionProfileController.hpp
```diff
@@ -240,6 +243,14 @@ class AsyncMotionProfileController : public AsyncPositionController<std::string,
    */
   void loadPath(std::string idirectory, std::string ipathId);
 
+  /**
+   * Attempts to remove a path without stopping execution, then if that fails,
+   * disables the controller and removes the path
+   *
+   * @param ipathId The path ID that will be removed
+   */
+  void forceRemovePath(std::string ipathId);
```
Octogonapus: Should be a const ref

include/okapi/api/coreProsAPI.hpp
```diff
@@ -94,3 +98,29 @@ class CrossplatformThread {
...
+#ifdef THREADS_STD
+    mutex.unlock();
+#else
+    mutex.give();
+#endif
+  }
+
+  protected:
+  CROSSPLATFORM_MUTEX_T mutex;
+};
```
Octogonapus: This file should end with a newline.

include/okapi/api/coreProsAPI.hpp
```diff
@@ -94,3 +98,29 @@ class CrossplatformThread {
 
   CROSSPLATFORM_THREAD_T thread;
 };
+
+class CrossplatformMutex {
+  public:
+  CrossplatformMutex() {
```
Octogonapus: How can we use RAII to guarantee the lock is always released, even in the case of an unexpected exception? Just calling `unlock()` isn't enough because the method could exit early. [std::lock_guard](https://en.cppreference.com/w/cpp/thread/lock_guard) uses the constructor to take the lock and the destructor to release it. Can we do the same here? We still need an `unlock()` method though because we need to be able to release it early (i.e., before destruction).

src/api/control/async/asyncLinearMotionProfileController.cpp
```diff
@@ -135,12 +135,26 @@ AsyncLinearMotionProfileController::getPathErrorMessage(const std::vector<Waypoi
...
+
   auto oldPath = paths.find(ipathId);
   if (oldPath != paths.end()) {
     free(oldPath->second.segment);
     paths.erase(ipathId);
   }
+
+  pathRemoveMutex.unlock();
+
+  // A return value of true mirrors the previous behavior of removePath
```
Octogonapus: Don't talk about the previous behavior here because nobody will remember it.

src/api/control/async/asyncLinearMotionProfileController.cpp
```diff
@@ -315,4 +338,15 @@ CrossplatformThread *AsyncLinearMotionProfileController::getThread() const {
 
 void AsyncLinearMotionProfileController::tarePosition() {
 }
+
+void AsyncLinearMotionProfileController::forceRemovePath(std::string ipathId) {
+  if (!removePath(ipathId)) {
+    std::string message =
+      "AsyncLinearMotionProfileController: Disabling controller to remove path " + ipathId;
+    LOG_WARN(message);
```
Octogonapus: Inline variable `message`.

src/api/control/async/asyncLinearMotionProfileController.cpp
```diff
@@ -315,4 +338,15 @@ CrossplatformThread *AsyncLinearMotionProfileController::getThread() const {
...
+    this->flipDisable(true);
```
Octogonapus: Remove `this->`.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -175,13 +175,26 @@ std::string AsyncMotionProfileController::getPathErrorMessage(const std::vector<
...
   auto oldPath = paths.find(ipathId);
   if (oldPath != paths.end()) {
     free(oldPath->second.left);
     free(oldPath->second.right);
     paths.erase(ipathId);
   }
+
+  pathRemoveMutex.unlock();
+
+  // A return value of true mirrors the previous behavior of removePath
```
Octogonapus: Don't talk about the previous behavior here because nobody will remember it.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -497,4 +512,14 @@ std::string AsyncMotionProfileController::makeFilePath(std::string directory,
 
   return path;
 }
+
+void AsyncMotionProfileController::forceRemovePath(std::string ipathId) {
+  if (!removePath(ipathId)) {
+    std::string message =
+      "AsyncMotionProfileController: Disabling controller to remove path " + ipathId;
+    LOG_WARN(message);
```
Octogonapus: Inline variable `message`.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -497,4 +512,14 @@ std::string AsyncMotionProfileController::makeFilePath(std::string directory,
...
+    this->flipDisable(true);
```
Octogonapus: Remove `this->`.

Octogonapus (changes requested):
This looks good. The only major thing is my comment on `CrossplatformMutex` RAII.

include/okapi/api/coreProsAPI.hpp
juliaschatz: Turns out we actually get to use lock_guard with almost no changes.

Octogonapus (commented):
I will test this on a real robot tomorrow.

Octogonapus (approved):

Octogonapus:
@nickschatz Great work on this, thanks for spending the time.

PR #380, opened 2019-08-23
Merge odometry branch

### LINKED ISSUE #213: Finish odometry
Odometry has been sitting around in a "basically done we just need to test it for reals this time" state. This is an issue so we can track it officially.

Octogonapus:
### Description of the Change

This PR merges the odometry feature set. This is being merged in an RC-ready condition. This PR also removes the integrated PID-related function from AbstractMotor and changes the ADI classes to use PROS' C API.

Note: This PR is not labeled as `needs docs` because the documentation for v4 is being moved to Doxygen, which generates documentation from header files. There is a separate branch which improves the documentation for this feature set.

### Verification Process

I have done manual integration testing on this branch for a few months, as have a few community members. The latest version has been tested driving around a robot without any issues.

### Applicable Issues

Closes #213.


include/okapi/api/chassis/model/chassisModel.hpp
```diff
@@ -126,83 +126,31 @@ class ChassisModel : public ReadOnlyChassisModel {
   virtual void setGearing(AbstractMotor::gearset gearset) = 0;
 
   /**
-   * Sets new PID constants.
+   * Sets a new maximum velocity in RPM [0-600].
```
theol0403: How do we know the velocity range is [0-600]? Does okapi normalize it? Are we just taking the value of the largest possible gearset? If so, it might be better to mention that the RPM range depends on the gearset.

include/okapi/api/odometry/odomMath.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Computes the distance and angle from the given Odometry state to the given point.
+   *
+   * @param ix x coordinate
+   * @param iy y coordinate
+   * @param istate odometry state
+   * @return distance and angle to the point
+   */
+  static std::tuple<QLength, QAngle>
```
theol0403: Wouldn't this be nicer as an `std::pair`?

theol0403 (commented):
I read through some of it, it looks really good. I see a lot of great improvements and stuff I'm really happy to see.
I like that setting the constants for the motor's integrated PID is gone, as it was awkward and could cause problems.

include/okapi/api/chassis/controller/odomChassisController.hpp
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Drives the robot straight to a point in the odom frame.
+   *
+   * @param ix x coordinate
+   * @param iy y coordinate
+   * @param ibackwards whether to drive to the target point backwards
+   * @param ioffset offset from target point in the direction pointing towards the robot
+   */
+  virtual void driveToPoint(const QLength &ix,
+                            const QLength &iy,
```
theol0403: What `StateMode` do these values refer to? It might be a good idea to have a flag in the chassis controller and maybe a parameter in the ctor to specify what mode the entire chassis uses.

include/okapi/api/chassis/model/chassisModel.hpp
Octogonapus: It's RPM. The maximum does vary based on the gearset, yes, so if you set `600` max but actually have `200` max and tell the chassis to go full velocity, it will try to go `600`. That could be mentioned in the docs.

include/okapi/api/odometry/odomMath.hpp
Octogonapus: Probably.

theol0403 (commented):
I looked at the odometry, I have a few opinions. 
#### Coordinates
I think that for all the coordinate parameters, which involve an `x` and a `y`, it should be refactored to a `Point` or `2dPoint`. This has a few benefits.  First, there is less boilerplate as only one parameter is needed. Second, the API is barely changed - `chassis.driveToPoint({x, y});`. Third, it allows people to store points in their code, and then use those points multiple times:
```cpp
Point usefulPoint{1_ft, 11_ft};
chassis.driveToPoint(usefulPoint);
```
I think this pattern should be used everywhere, from `OdomMath::computeAngleToPoint(Point)` to `OdomChassisController::driveToPoint(Point)`.
#### Additional movement functions
It would be nice if the following odom function could be added: `void OdomChassisController::turnToPoint(Point &ipoint);`

include/okapi/api/chassis/model/chassisModel.hpp
theol0403: I think it may be confusing to people, and they will assume that regardless of their actual gearset the range will be `[0, 600]`. It's not informative anyways, because it assumes people will already know what their gearset is, disregarding what the comment says.
I think that should be taken out entirely, and replaced with either of the following:
```cpp
/**
 * Sets a new maximum velocity in RPM, according to the gearset.
 * Sets a new maximum velocity in RPM.
```

include/okapi/api/chassis/controller/odomChassisController.hpp
Octogonapus: That parameter should be added, this is an oversight.

src/impl/device/motor/motorGroup.cpp
```diff
@@ -9,6 +9,21 @@
...
+  if (imotors.size() == 0) {
+    std::string msg(
+      "MotorGroup: A MotorGroup must be created with at least one motor. No motors were given.");
+    LOG_ERROR(msg);
+    throw std::invalid_argument(msg);
+  }
+
+  motors.reserve(imotors.size());
+  for (auto motor : imotors) {
+    motors.push_back(std::make_shared<Motor>(motor));
```
theol0403, line 22: I could be mistaken, but I believe in the `MotorGroup` ctor, the copy constructor of the input motors is called twice. 
It gets first copied when the motors get passed through the `initializer_list`, it then gets copied again when `make_shared` allocates new memory. I might be being silly and this is negligible, but I think it might be a good idea to pass the motors through the initializer list by universal reference.

Octogonapus: I think it's just once when the shared_ptr is made. We need to make at least one copy, so I am not that concerned.

theol0403 (commented):
Thanks for this. 
If I were using this API to write an autonomous program, and I wanted to visualize the field in cartesian, I think it would be awkward to have to put the StateMode in every single call:
```cpp
chassis.driveToPoint({0_ft, 5_ft}, StateMode::CARTESIAN);
chassis.driveToPoint({1_ft, 3_ft}, StateMode::CARTESIAN);
chassis.driveToPoint({5_ft, 7_ft}, StateMode::CARTESIAN);
```
Instead, it would be nicer to have a chassis-wide flag like `setTurnsMirrored` that we can set once.
It could be modified through a setter method, but it could also be through a parameter in the ctor.
People will want to choose one mindset and stick with it, so it is redundant to ask for it in every single command.

src/api/odometry/odomMath.cpp
```diff
@@ -14,24 +14,47 @@ OdomMath::OdomMath() = default;
...
 OdomMath::~OdomMath() = default;
 
-QLength OdomMath::computeDistanceToPoint(const Point2d &ipoint, const OdomState &istate) {
-  const double xDiff = (ipoint.x - istate.x).convert(meter);
-  const double yDiff = (ipoint.y - istate.y).convert(meter);
-  return std::sqrt((xDiff * xDiff) + (yDiff * yDiff)) * meter;
+QLength OdomMath::computeDistanceToPoint(const Point2d &ipoint,
+                                         const OdomState &istate,
+                                         const StateMode &ipointMode) {
+  if (ipointMode == StateMode::CARTESIAN) {
```
theol0403: This is a little bit redundant. How about:
```cpp
const double xDiff = (ipoint.inMode(ipointMode).y - istate.x).convert(meter);
const double yDiff = (ipoint.inMode(ipointMode).x - istate.y).convert(meter);
return std::sqrt((xDiff * xDiff) + (yDiff * yDiff)) * meter;
```
The same goes for `computeAngleToPoint` and `computeDistanceAndAngleToPoint`

src/api/odometry/odomMath.cpp
```diff
@@ -14,24 +14,47 @@ OdomMath::OdomMath() = default;
...
+    const double yDiff = (ipoint.x - istate.y).convert(meter);
+    return (std::atan2(yDiff, xDiff) * radian) - istate.theta;
+  } else {
+    const double xDiff = (ipoint.x - istate.x).convert(meter);
+    const double yDiff = (ipoint.y - istate.y).convert(meter);
+    return (std::atan2(yDiff, xDiff) * radian) - istate.theta;
+  }
 }
 
 std::pair<QLength, QAngle> OdomMath::computeDistanceAndAngleToPoint(const Point2d &ipoint,
```
theol0403: I think `computeDistanceAndAngleToPoint` should delegate:
```cpp
return std::make_pair(computeDistanceToPoint(ipoint, istate, ipointMode), computeAngleToPoint(ipoint, istate, ipointMode));
```

src/api/odometry/odomMath.cpp
Octogonapus: I am still working on it.

src/api/odometry/odomMath.cpp
Octogonapus: It doesn't delegate to save computing the diff's twice.

include/okapi/api/chassis/controller/chassisController.hpp
```diff
@@ -92,213 +90,36 @@ class ChassisController : public ChassisModel {
...
-  void stop() override;
-
-  /**
-   * Drive the robot with a tank drive layout. Uses voltage mode.
-   *
-   * @param ileftSpeed left side speed
-   * @param irightSpeed right side speed
-   * @param ithreshold deadband on joystick values
-   */
-  void tank(double ileftSpeed, double irightSpeed, double ithreshold = 0) override;
```
baylessj, line 150: I'm assuming that the removal of these functions indicates that the user should call these from the internal ChassisModel, right? This change would break some of my code from last season but as long as it is still possible to have full control over the underlying ChassisModel then it would be find to work around this change.

include/okapi/api/chassis/controller/defaultOdomChassisController.hpp
```diff
@@ -0,0 +1,110 @@
...
+                   const StateMode &imode = StateMode::FRAME_TRANSFORMATION) override;
+
+  /**
+   * Turns the robot to face an angle in the odom frame.
+   *
+   * @param iangle The angle to turn to.
+   */
+  void turnToAngle(const QAngle &iangle) override;
+
+  void moveDistance(QLength itarget) override;
```
baylessj: Is the plan to add docstrings for these at some point or are the going to be left as-is because they're used/documented in other classes?

src/api/chassis/controller/defaultOdomChassisController.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+void DefaultOdomChassisController::turnToPoint(const Point &ipoint, const StateMode &imode) {
+  const auto angle =
+    OdomMath::computeAngleToPoint(ipoint, odom->getState(StateMode::FRAME_TRANSFORMATION), imode);
+
+  LOG_INFO("DefaultOdomChassisController: Computed angle of " +
+           std::to_string(angle.convert(degree)) + " degrees");
+
+  if (angle.abs() > turnThreshold) {
```
baylessj, line 60: I'm not sure if it would be appropriate to have the default behavior be to have a non-zero `turnThreshold` and not move small amounts. I could see users being confused by this behavior, and I've never encountered a situation where I wanted the robot to not move in the case of a tiny adjustment.

baylessj: It makes sense to have a turnThreshold for odometry calculation for filtering purposes, but I don't think that same logic should carry over to this.

baylessj (changes requested):
Glad to see these changes getting merged in, there's a lot of really good functionality here. My primary concern with the changes is the same as @theol0403, I'd like to see a global switch for Cartesian coordinates rather than passing in a switch to every relevant function.

I noticed that there are a number of somewhat unrelated changes being grouped into this PR such as a switch to using PROS C types internally and removing the ability to modify the motors' PID values. Is there a notice of this in the other PRs that are being merged in with this, or could this be added to the description of this PR? I think those are positive changes, I just would like to be able to see where those changes were made when looking back at revision history without too much digging.

include/okapi/api/chassis/controller/chassisController.hpp
Octogonapus: Users can call these from the internal ChassisModel, yes.

src/api/chassis/controller/defaultOdomChassisController.cpp
Octogonapus: I will make them zero by default, but they were very important for the autonomous robot. Most of this code started with that bot.

baylessj: I could see where that would be important in the case where trajectories are automatically generated, that makes sense. I like the idea to set their defaults to zero, that seems like a good compromise to me :+1:

baylessj (approved):
Looks like my requested changes have been applied, this all looks good to me :+1:

PR #388, opened 2019-10-10
Fix issue #387 controllerSet() on reversed Motors

### LINKED ISSUE #387: controllerSet() doesn't respect motor being reversed
**Describe the bug**
On a reversed motor, `moveRelative()` will be reversed but not `controllerSet()`.

**Steps To Reproduce**
```
okapi::Motor myMotor(-5);
myMotor.setGearing(okapi::AbstractMotor::gearset::green);
//Turns at 200 RPM in reverse.
myMotor.moveVelocity(200);
pros::delay(1000);
//Turns at 200 RPM forwards.
myMotor.controllerSet(1);
pros::delay(1000);
```

**Expected behavior**
I expected that, on a reversed motor, `controllerSet(1);` will move it in reverse.

**Versions**
```
PROS Project for v5 at: /home/ungato/strelizia (strelizia)
Name      Version                 Origin
--------  ----------------------  --------
kernel    3.1.7-dirty.35.fde44c4  local
okapilib  4.0.0-RC2               local
```

**Additional Information**
The issue is caused by the `reversed` field being used twice, once by `controllerSet()` and again by `moveVelocity()` which it calls.

vexcat:
### Description of the Change

Remove the reference to `reversed` in Motor's `controllerSet`, as the `moveVelocity` function it calls already uses it.

### Motivation

Doing this allows for controllers to behave as expected on a reversed motor.

### Possible Drawbacks

Anyone working around this issue will have their motors again doubly-reversed, though it should be fine as not many people have the RC2 version of OkapiLib.

### Applicable Issues

Closes #387.

### DIFF
--- a/src/impl/device/motor/motor.cpp
+++ b/src/impl/device/motor/motor.cpp
@@ -280,7 +280,7 @@ std::shared_ptr<ContinuousRotarySensor> Motor::getEncoder() {
 }
 
 void Motor::controllerSet(const double ivalue) {
-  //Reversing is done by moveVelocity, no need to add * reversed.
+  // Reversing is done by moveVelocity, no need to reverse here
   moveVelocity(ivalue * toUnderlyingType(getGearing()));
 }
 
--- a/src/impl/device/motor/motor.cpp
+++ b/src/impl/device/motor/motor.cpp
@@ -280,7 +280,8 @@ std::shared_ptr<ContinuousRotarySensor> Motor::getEncoder() {
 }
 
 void Motor::controllerSet(const double ivalue) {
-  moveVelocity(ivalue * toUnderlyingType(getGearing()) * reversed);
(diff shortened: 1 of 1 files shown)


src/impl/device/motor/motor.cpp
```diff
@@ -280,7 +280,8 @@ std::shared_ptr<ContinuousRotarySensor> Motor::getEncoder() {
 }
 
 void Motor::controllerSet(const double ivalue) {
-  moveVelocity(ivalue * toUnderlyingType(getGearing()) * reversed);
+  //Reversing is done by moveVelocity, no need to add * reversed.
```
Octogonapus:
```suggestion
  // Reversing is done by moveVelocity, no need to reverse here
```

Octogonapus (changes requested):

PR #392, opened 2019-10-24
Improve v4 docs

theol0403:
### Description of the Change

Here are some changes that I think improves the docs in my eyes.

- Add namespace instructions to Getting Started
- Fix Clawbot Tutorial voltage movement commands
- Clean up, fix, and slightly rework Clawbot Tutorial
- Remove unnecessary motor port constants in cases where they were obvious and bloated the code
- Favor Autonomous Movement driving using units

### Other Suggestions
Here are a few suggestions I recommend adding to the docs, but I am unsure how to do so:

- Replace "Pages" tab in header with "Tutorials" tab linking to tutorials
- Add "API" tab in header between "Tutorials" and "Classes", which lists and organizes the classes in a similar way to the PROS docs. This organization was very helpful to me when learning okapilib.
- Remove the first-line indent at the beginning of each paragraph. This decreases readability and wastes space, especially in 1-word sentences.
- In the "Public functions" section of each page, increase line wrap to improve compactness and readability 
- In the "Public functions" section, match function signature to actual prototype instead of the unfamiliar `auto get() const -> double` syntax, as opposed to `double okapi::ADIGyro::get() const`


theol0403:
I was wrong about having the ADIButton only accept an integer, I forgot about the character transform.

theol0403:
Sorry, I must have tired when I made this. I realize that in those examples, chassis scales are not provided, so driving using units won't work.

Octogonapus:
> 
> 
> Sorry, I must have tired when I made this. I realize that in those examples, chassis scales are not provided, so driving using units won't work.

They should be provided then.

Octogonapus:
> Replace "Pages" tab in header with "Tutorials" tab linking to tutorials

I would like to keep pages, but I'm fine with also having "Tutorials". Not sure if that's easy to do, though. The "tabs" you're referring to are selected from a predefined set of options.

> Add "API" tab in header between "Tutorials" and "Classes", which lists and organizes the classes in a similar way to the PROS docs. This organization was very helpful to me when learning okapilib.

We can definitely add this as a page.

> Remove the first-line indent at the beginning of each paragraph. This decreases readability and wastes space, especially in 1-word sentences.
In the "Public functions" section of each page, increase line wrap to improve compactness and readability
In the "Public functions" section, match function signature to actual prototype instead of the unfamiliar auto get() const -> double syntax, as opposed to double okapi::ADIGyro::get() const

Not sure if these are configurable. The Doxygen processor we're currently using is very opinionated.

theol0403:
> Not sure if that's easy to do, though. The "tabs" you're referring to are selected from a predefined set of options.

Looks like this is possible [here](https://mcss.mosra.cz/documentation/doxygen/#navbar-links).
We could also use the submenu functionality to implement a dropdown under "API" which contains each API section, I think that would be quite nice.

theol0403:
> Remove the first-line indent at the beginning of each paragraph. This decreases readability and wastes space, especially in 1-word sentences.

I think I've figured out how to do this.
The indent is specified here:
https://github.com/mosra/m.css/blob/2160a24dd96c3edb8ddf22ce79838df2fba40235/css/m-theme-dark.css#L32
So we can either:
 - Ship custom theme file with okapi
 - Inject CSS into the output, I believe there is an API for that using the Doxyfile

Removing the indent in my browser's inspector made it look so much better 👌

theol0403:
> In the "Public functions" section of each page, increase line wrap to improve compactness and readability

Changing the wrap in my browser inspector makes it look much better.

From:

To:

theol0403:
There are some problems with disabling the wrap. They show in ChassisControlleBuilder for example, basically they remove needed line breaks.
Extending the width of the entire content should have the same intended effect.

theol0403:
I've completed the list.
However, I've noticed another issue, which would be great if @Octogonapus could look into:

mathUtil.hpp is not being indexed due to not being a class. There should be some docstring flag you can add to make it index the file.

Octogonapus:
> mathUtil.hpp is not being indexed due to not being a class. There should be some docstring flag you can add to make it index the file.

Its contents are still being documented (meaning they are searchable and can be referred to). I added some docs for them on the main branch and merged your branch up.

PR #394, opened 2019-11-01
Add HDriveModel

Octogonapus:
### Description of the Change

This PR adds support for h-drives via `HDriveModel`.

### Motivation

A user was asking for this recently.

### Possible Drawbacks

None.

### Verification Process

Unit tests only. I have not verified this on real hardware.

### Applicable Issues

None.

### DIFF
--- a/src/api/chassis/model/hDriveModel.cpp
+++ b/src/api/chassis/model/hDriveModel.cpp
@@ -6,6 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 #include "okapi/api/chassis/model/hDriveModel.hpp"
+#include "okapi/api/util/mathUtil.hpp"
 #include <algorithm>
 #include <utility>
 
@@ -226,13 +227,7 @@ double HDriveModel::getMaxVelocity() const {
 }
 
 void HDriveModel::setMaxVoltage(double imaxVoltage) {
-  if (imaxVoltage < 0) {
-    maxVoltage = 0;
-  } else if (imaxVoltage > 12000) {
-    maxVoltage = 12000;
-  } else {
-    maxVoltage = imaxVoltage;
-  }
+  maxVoltage = std::clamp(imaxVoltage, 0.0, v5MotorMaxVoltage);
 }
 
 double HDriveModel::getMaxVoltage() const {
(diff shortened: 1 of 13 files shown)


Octogonapus:
@theol0403 Could you please take a look over this?

Octogonapus:
Anyone looking to test this, the artifact is here: [template.zip](https://github.com/OkapiLib/OkapiLib/files/3796370/template.zip)

Sleuth56:
I've tested it. It works fine. It might be good to add a third controller input to the arcade drive. I got around it with a separate line controlling the middle motor.
`    driveTrain->arcade(controller.getAnalog(ControllerAnalog::leftY),controller.getAnalog(ControllerAnalog::rightX), 0.3);
    driveTrain->middle(controller.getAnalog(ControllerAnalog::leftX));`

Octogonapus:
> It might be good to add a third controller input to the arcade drive

Use `hArcade` for that. `arcade` is an overridden method.

theol0403:
@lerker100 
I'm guessing that you don't know how to use `hArcade` from a `ChassisController`.
The answer is that you need to get the model from the chassis (which is internally an `HDriveModel`, but gets returned as a generic `ChassisModel`), and then cast (tell the compiler what *type* of model you have) the `ChassisModel` into a `HDriveModel`.
For an example doing the same thing with an x-drive, check this out:
https://www.vexforum.com/t/xdriver-model-does-not-have-xarcade/67778/4?u=theol0403

src/api/chassis/model/hDriveModel.cpp
```diff
@@ -0,0 +1,247 @@
...
+  } else {
+    maxVelocity = imaxVelocity;
+  }
+}
+
+double HDriveModel::getMaxVelocity() const {
+  return maxVelocity;
+}
+
+void HDriveModel::setMaxVoltage(double imaxVoltage) {
```
theol0403, line 229: I'm not sure if this is the same thing for all the models, but this function can be simplified to
```cpp
maxVoltage = std::clamp(imaxVoltage, 0, 12000);

**Review by theol0403 — CHANGES_REQUESTED** [review]
In functions that do not use the middle wheel, I recommend stopping the middle wheel. 
That way, doing 
```cpp
hArcade(1, 1, 0);
delay(200);
rotate(1);
delay(500);
```
won't make the robot try to strafe while rotating.

The rest looks good to me, though I did not fully go through the builder logic.

theol0403 (approved):

Sleuth56:
Yea I found it after I said that but wasn't able to test it on hardware yet

PR #396, opened 2019-11-02
Add odometry for all chassis models

Octogonapus:
### Description of the Change

This PR adds odometry support for all chassis models. This PR also adds `ThreeEncoderXDriveModel`.

### Motivation

This is a good feature to have.

### Possible Drawbacks

None.

### Verification Process

I have not verified this on real hardware.

### Applicable Issues

None.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ add_executable(OkapiLibV5
         include/okapi/api/chassis/model/readOnlyChassisModel.hpp
         include/okapi/api/chassis/model/skidSteerModel.hpp
         include/okapi/api/chassis/model/threeEncoderSkidSteerModel.hpp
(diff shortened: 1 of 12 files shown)


Octogonapus:
@theol0403 Let me know how your testing goes.

theol0403:
There is a bigger problem I have found when using ChassisControllerBuilder odom, and CCI.
When you specify chassis width (and gearset), what does that apply to?
The math for CCI? Or the math for odom?
Some refactor is needed to disambiguate this.

As far as testing, it seems like the odometry functioned, but the movement was wrong (odom scales being applied to CCI).

Octogonapus:
@theol0403 Fixing that issue has moved to https://github.com/OkapiLib/OkapiLib/pull/397

PR #426, opened 2020-01-24
#422: Initialize controller buttons

### LINKED ISSUE #422: Uninitialized controller buttons
**Describe the bug**
Accessing a controller button from a controller that was constructed non-globally by means of `operator[](ControllerDigital ibtn)` causes a segfault. 

This is because the `buttonArray` member of `Controller` is not guaranteed to be zero-initialized, so in some cases the nullptr check wrongly fails: 
https://github.com/OkapiLib/OkapiLib/blob/f025ccaafd4da81319759631e254b55dde096afa/include/okapi/impl/device/controller.hpp#L117 https://github.com/OkapiLib/OkapiLib/blob/f025ccaafd4da81319759631e254b55dde096afa/src/impl/device/controller.cpp#L42

This causes a segfault because the pointer is never lazy-initialized. 
Creating the controller globally avoids the problem, since there is a good chance global variables contain 0.

**Steps To Reproduce**
```cpp
void opcontrol() {
  Controller controller;
  auto button = controller[ControllerDigital::R1];
}
```

**Expected behavior**
Not a segfault.

**Versions**
Okapilib 4.0.2

**Additional Information**
Discord conversation on possible solutions: https://canary.discordapp.com/channels/197777408198180864/198658294007463936/665994944166494218

theol0403:
### Description of the Change

This PR fixes #422.
Following the rules of aggregate initialization, an empty initializer list will zero-initialize all elements of the array.

### Motivation

Having an uninitialized array can lead to segfaults.

### Verification Process

- [ ] Test using the following code:
```cpp
void opcontrol() {
  Controller controller;
  auto button = controller[ControllerDigital::R1];
}
```

I can probably test when I have the robot on monday.

### DIFF
--- a/include/okapi/impl/device/controller.hpp
+++ b/include/okapi/impl/device/controller.hpp
@@ -114,6 +114,6 @@ class Controller {
   protected:
   ControllerId okapiId;
   pros::controller_id_e_t prosId;
-  std::array<ControllerButton *, 12> buttonArray {};
+  std::array<ControllerButton *, 12> buttonArray{};
 };
 } // namespace okapi
--- a/include/okapi/impl/device/controller.hpp
+++ b/include/okapi/impl/device/controller.hpp
@@ -114,6 +114,6 @@ class Controller {
   protected:
   ControllerId okapiId;
   pros::controller_id_e_t prosId;
-  std::array<ControllerButton *, 12> buttonArray;
+  std::array<ControllerButton *, 12> buttonArray {};
 };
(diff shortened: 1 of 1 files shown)


Octogonapus (approved):

theol0403:
I reproduced the issue and verified that this PR resolves it.

I had to run this to reproduce the issue:
```cpp
Controller& controller = *(new Controller);
auto btn = controller[ControllerDigital::R1];
```

PR #427, opened 2020-01-24
Fix odometry docs when using CCPID

theol0403:
### Description of the Change

The odometry docs show that the scales passed to `withDimensions` are for the powered wheels when using external sensors and CCPID. This is wrong, there should only be tracking wheel scales in this scenario.

I also changed the tracking wheel diameter in the examples to be more realistic.

### Motivation

People may have got confused.

### Verification Process

Read it.


Octogonapus (approved):

Octogonapus:
Because these docs changes work with the latest release, I am fine merging these directly into `master` to get them out faster.

PR #436, opened 2020-02-17
Fixed Pathfinder as a C++ library

### LINKED ISSUE #435: Linking fails with only Pathfinder library
**Describe the bug**
Compiler can't seem to find Pathfinder library, but can find adiGyro just fine
`
/opt/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: bin/autonomous.cpp.o: in function pathfinder_test():
/home/ryan/workspaces/git/2019-2020-BIG/src/autonomous.cpp:90: undefined reference to d2r(double)
`

Same error message for `pathfinder_prepare`, `pathfinder_generate`, `pathfinder_modify_tank`, `pathfinder_follow_encoder`, and `pf_fit_hermite_cubic`

**Steps To Reproduce**
Call any one of the functions above

**Versions**
Pros 3.2.1 and Okapilib 4.0.3

**Additional Information**
Checked inside okapilib.a and the object files do exist for pathfinder, gcc is just unable to see them?

superrm11:
### Description of the Change

Because pathfinder was built as a C library, in order to be used with C++ pros projects, it must be labeled as such. Therefore extern "C" was added to the header files.

### Motivation

Pathfinder failed to link while compiling a pros project for the RIT  VEXU team

### Possible Drawbacks

N/A

### Verification Process

Ran pathfinder_generate(...), pathfinder_modify_tank(...), dtr(...) and pathfinder_follow_encoder(...) without errors occurring

### Applicable Issues

Closes #435.

### DIFF
--- a/include/okapi/pathfinder/include/pathfinder/fit.h
+++ b/include/okapi/pathfinder/include/pathfinder/fit.h
@@ -1,6 +1,11 @@
 #ifndef PATHFINDER_FIT_H_DEF
 #define PATHFINDER_FIT_H_DEF
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 #include "okapi/pathfinder/include/pathfinder/lib.h"
 #include "okapi/pathfinder/include/pathfinder/structs.h"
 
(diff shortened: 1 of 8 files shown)


Octogonapus:
Ok I think I finally understand your issue: you are trying to use Pathfinder's API directly, instead of using one of OkapiLib's classes, which I guess nobody has done before because I have never heard a report like this.

Octogonapus (approved):

PR #441, opened 2020-03-30
Minimize turns

### LINKED ISSUE #439: Minimize turns to a maximum of 180 degrees
**Describe the bug**
Per: https://www.vexforum.com/t/odometry-not-turning-the-correct-way/78109 it is possible for a turn to go 270 degrees clockwise when a 90 degree counter-clockwise turn would accomplish the same thing.

**Steps To Reproduce**
RPM reproduced the behavior in the thread:

What it is doing is calculating atan of (about) -.4/-.3 which is -127 deg and then subtracting the current heading of 177 (ideally it would have been 180) from it (-127 -177) or about -304. Of course odom is using the end position from the encoders so the 0-.4 and .7-1 are the ideal number which is why the log shows -307. I guess they could optimize it by looking to see if if adding 360 or subtracting 360 would result in an smaller magnitude (or any other method). 

**Expected behavior**
Per above, "they could optimize it by looking to see if if adding 360 or subtracting 360 would result in an smaller magnitude (or any other method). "

**Versions**
Name      Version    Origin
--------  ---------  -------------
kernel    3.2.1      pros-mainline
okapilib  4.0.3      pros-mainline

**Additional Information**
Though it might make more sense to implement this at the ChassisController level

dcieslak19973:
### Description of the Change

Changed the odometry chassis controller to minimize turns to 180 degrees (either clockwise or counterclockwise)
### Motivation

While it was suggested to do this in the chassisController, the unit tests there assumed full control over the chassis, so it made more sense to do this in the odometry chassis controller.

### Possible Drawbacks

Favors counterclockwise turns of 180 degrees.

### Verification Process

New unit tests were created to validate the functionality

### Applicable Issues

Closes #439


src/api/odometry/odomMath.cpp
```diff
@@ -44,4 +44,13 @@ double OdomMath::computeDistance(double xDiff, double yDiff) {
 double OdomMath::computeAngle(double xDiff, double yDiff, double theta) {
   return std::atan2(yDiff, xDiff) - theta;
 }
+
+double OdomMath::constrainAngle360(double theta) {
```
Octogonapus: I would rather these new methods use the units API instead of assuming that `theta` is in degrees. Then you can also avoid calling the `QAngle` constructor above.

Octogonapus (changes requested):

src/api/odometry/odomMath.cpp
dcieslak19973: Changed doubles to QAngle. Not sure the unit tests are terribly elegant now, but they are serviceable.

src/api/odometry/odomMath.cpp
```diff
@@ -44,4 +44,15 @@ double OdomMath::computeDistance(double xDiff, double yDiff) {
 double OdomMath::computeAngle(double xDiff, double yDiff, double theta) {
   return std::atan2(yDiff, xDiff) - theta;
 }
+
+QAngle OdomMath::constrainAngle360(QAngle theta) {
+  double degreeValue = theta.convert(degree) - 360.0 * std::floor(theta.convert(degree) / 360.0);
+  return QAngle(degreeValue * degreeToRadian);
```
Octogonapus: Please don't rely on `QAngle` storing its internal value in radians. You can just do `degreeValue * degree`. (here and below)

Octogonapus (changes requested):

Octogonapus:
Also, you should update with `develop` since you originally branched off of `master`.

src/api/odometry/odomMath.cpp
dcieslak19973: Done

dcieslak19973:
> Also, you should update with `develop` since you originally branched off of `master`.

Done; apologies for doing the merge in the same commit as above.

Octogonapus (approved):
Something weird happened in your merge, so I fixed things up and pushed to your branch.

PR #445, opened 2020-07-06
Issue #302: Add common math functions to RQuantity

### LINKED ISSUE #302: Add common math functions to RQuantity
The units API can be verbose to use when you need common math functions such as `sqrt()`, `sin()`, `cos()`, etc. `RQuantity` should define these.

breqdev:
### Description of the Change

Add common math and trig functions (`sin()`, `cos()`, `sqrt()`, etc.) to RQuantity (units API). This PR includes many of the functions present in the <cmath> header (trigonometric and hyperbolic functions, a few basic root functions, modulus, etc). It does not include rounding functions or log/exponent functions.

### Motivation

Previously, using common math functions with the units API required converting a number from an RQuantity to a double, using the function, and then converting the number back to an RQuantity. This PR adds functions like `sin()` that take RQuantity objects as arguments so users do not need to do this conversion.

### Possible Drawbacks

Because these functions are implemented as free functions [as per this comment](https://github.com/OkapiLib/OkapiLib/issues/302#issuecomment-619711767), they might clutter up the namespace a bit.

### Verification Process

I tested each function I implemented on my desktop to ensure it gave valid results. (I haven't gotten the chance to test on a V5 brain, but compilation succeeded when I copied these changes into a new PROS project. Since most of the work done by the units API is at compile time, this seemed like a good sign.)

### Applicable Issues

Closes #302.

### DIFF
--- a/include/okapi/api/units/RQuantity.hpp
+++ b/include/okapi/api/units/RQuantity.hpp
@@ -234,7 +234,7 @@ root(const RQuantity<M, L, T, A> &lhs) {
   return RQuantity<std::ratio_divide<M, std::ratio<R>>,
                    std::ratio_divide<L, std::ratio<R>>,
                    std::ratio_divide<T, std::ratio<R>>,
-                   std::ratio_divide<A, std::ratio<R>>>(std::pow(lhs.getValue(), 1/R));
+                   std::ratio_divide<A, std::ratio<R>>>(std::pow(lhs.getValue(), 1.0/R));
 }
 
 template <typename M, typename L, typename T, typename A>
--- a/include/okapi/api/units/RQuantity.hpp
+++ b/include/okapi/api/units/RQuantity.hpp
@@ -201,6 +201,18 @@ constexpr RQuantity<M, L, T, A> abs(const RQuantity<M, L, T, A> &rhs) {
   return RQuantity<M, L, T, A>(std::abs(rhs.getValue()));
 }
 
+template <typename R, typename M, typename L, typename T, typename A>
+constexpr RQuantity<std::ratio_multiply<M, R>,
+                    std::ratio_multiply<L, R>,
+                    std::ratio_multiply<T, R>,
+                    std::ratio_multiply<A, R>>
+pow(const RQuantity<M, L, T, A> &lhs) {
+  return RQuantity<std::ratio_multiply<M, R>,
+                   std::ratio_multiply<L, R>,
+                   std::ratio_multiply<T, R>,
+                   std::ratio_multiply<A, R>>(std::pow(lhs.getValue(), double(R::num) / R::den));
+}
+
 template <typename M, typename L, typename T, typename A>
 constexpr RQuantity<std::ratio_divide<M, std::ratio<2>>,
                     std::ratio_divide<L, std::ratio<2>>,
@@ -226,6 +238,30 @@ cbrt(const RQuantity<M, L, T, A> &rhs) {
 }
 
 template <typename M, typename L, typename T, typename A>
+constexpr RQuantity<std::ratio_multiply<M, std::ratio<2>>,
+                    std::ratio_multiply<L, std::ratio<2>>,
+                    std::ratio_multiply<T, std::ratio<2>>,
+                    std::ratio_multiply<A, std::ratio<2>>>
+square(const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<std::ratio_multiply<M, std::ratio<2>>,
+                   std::ratio_multiply<L, std::ratio<2>>,
+                   std::ratio_multiply<T, std::ratio<2>>,
+                   std::ratio_multiply<A, std::ratio<2>>>(std::pow(rhs.getValue(), 2));
+}
+
+template <typename M, typename L, typename T, typename A>
+constexpr RQuantity<std::ratio_multiply<M, std::ratio<3>>,
+                    std::ratio_multiply<L, std::ratio<3>>,
+                    std::ratio_multiply<T, std::ratio<3>>,
+                    std::ratio_multiply<A, std::ratio<3>>>
+cube(const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<std::ratio_multiply<M, std::ratio<3>>,
+                   std::ratio_multiply<L, std::ratio<3>>,
+                   std::ratio_multiply<T, std::ratio<3>>,
+                   std::ratio_multiply<A, std::ratio<3>>>(std::pow(rhs.getValue(), 3));
+}
+
+template <typename M, typename L, typename T, typename A>
 constexpr RQuantity<M, L, T, A> hypot(const RQuantity<M, L, T, A> &lhs,
                                       const RQuantity<M, L, T, A> &rhs) {
   return RQuantity<M, L, T, A>(std::hypot(lhs.getValue(), rhs.getValue()));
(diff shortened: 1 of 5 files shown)


breqdev:
It looks like the build failed. (I didn't think to run a `make all` before committing, because it didn't seem like these changes would have any effect on other parts of the code.)

I think this might be because the C `sin()`, `cos()`, etc. defined in `math.h` can't be overloaded? Several places in the OkapiLib code use these functions instead of `std::sin()` etc. defined in `<cmath>`.

I'm honestly not sure how best to proceed from here. I could put the functions into their own namespace, or make them member functions of RQuantity, but that doesn't seem ideal.

include/okapi/api/units/RQuantity.hpp
```diff
@@ -192,6 +193,157 @@ constexpr bool operator>(const RQuantity<M, L, T, A> &lhs, const RQuantity<M, L,
...
+constexpr RQuantity<M, L, T, std::ratio_subtract<A, std::ratio<1>>>
+cos(const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<M, L, T, std::ratio_subtract<A, std::ratio<1>>>(std::cos(rhs.getValue()));
+}
+
+template <typename M, typename L, typename T, typename A>
+constexpr RQuantity<M, L, T, std::ratio_subtract<A, std::ratio<1>>>
+tan(const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<M, L, T, std::ratio_subtract<A, std::ratio<1>>>(std::tan(rhs.getValue()));
+}
```
Octogonapus: I don't think we should keep the units from the operand. These trig functions all return a unitless value so I think our API should reflect that.

include/okapi/api/units/RQuantity.hpp
```diff
@@ -192,6 +193,157 @@ constexpr bool operator>(const RQuantity<M, L, T, A> &lhs, const RQuantity<M, L,
...
+constexpr RQuantity<M, L, T, std::ratio_add<A, std::ratio<1>>>
+acosh(const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<M, L, T, std::ratio_add<A, std::ratio<1>>>(std::acosh(rhs.getValue()));
+}
+
+template <typename M, typename L, typename T, typename A>
+constexpr RQuantity<M, L, T, std::ratio_add<A, std::ratio<1>>>
+atanh(const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<M, L, T, std::ratio_add<A, std::ratio<1>>>(std::atanh(rhs.getValue()));
+}
```
Octogonapus: These should return just an angle.

Octogonapus (changes requested):
Thanks for taking the time to do this PR. I just had some small comments.

@theol0403 Can you comment on what may be missing from this PR?

breqdev:
Thanks for the feedback. To clarify, if (for instance) `sin()` returns only a unitless number, should it accept only an angle? Also, if the trig functions only work with angles and numbers, would it make more sense to put them in `QAngle.hpp`, or to have all the functions together in `RQuantity.hpp`?

It looks like there are more files that use (e.g.) `sin()` instead of `std::sin()` that are preventing the build from passing. I'm happy to go through and fix these if you want, but I don't really have a way to test the result.

Octogonapus:
> To clarify, if (for instance) `sin()` returns only a unitless number, should it accept only an angle?

Yes.

> Also, if the trig functions only work with angles and numbers, would it make more sense to put them in `QAngle.hpp`, or to have all the functions together in `RQuantity.hpp`?

Leave them in `RQuantity.hpp`.

> I'm happy to go through and fix these if you want, but I don't really have a way to test the result.

Go ahead and change them.

Octogonapus (approved):

include/okapi/api/units/RQuantity.hpp
```diff
@@ -192,6 +193,151 @@ constexpr bool operator>(const RQuantity<M, L, T, A> &lhs, const RQuantity<M, L,
...
+                   std::ratio_divide<A, std::ratio<3>>>(std::cbrt(rhs.getValue()));
+}
+
+template <typename M, typename L, typename T, typename A>
+constexpr RQuantity<M, L, T, A> hypot(const RQuantity<M, L, T, A> &lhs,
+                                      const RQuantity<M, L, T, A> &rhs) {
+  return RQuantity<M, L, T, A>(std::hypot(lhs.getValue(), rhs.getValue()));
+}
+
+template <typename M, typename L, typename T, typename A>
```
theol0403, line 276: These methods should not be necessary. Internally, `std::max` uses comparison operators, which are compatible with RQuantity. This means the 
```cpp
template< class T > constexpr T max( std::initializer_list<T> ilist );
```
overload (which imo is quite useful) is also available.

theol0403 (approved):
Looks good to me, you've included all the useful functions.

I know you said it does not include rounding functions or log/exponent functions, but if possible those would be useful (especially pow). Pow would probably look different than the other methods, as the exponent would probably have to be given as a template parameter for it to know the return type. A good compromise would be a square and cube method.

Rounding methods like floor and ceil are also useful, but I can't think of a good way to express them in terms of units. An API such as `floor<degree>(0.6_deg)` might be possible, but that might belong to a separate PR.

Octogonapus:
> Pow would probably look different than the other methods, as the exponent would probably have to be given as a template parameter for it to know the return type. A good compromise would be a square and cube method.

I agree. @Breq16 can you try to implement pow this way?

> Rounding methods like floor and ceil are also useful, but I can't think of a good way to express them in terms of units

I agree they are useful and I don't think it would be difficult to implement. Wouldn't round, floor, and ceil simply return the same dimensions as they accepted?

include/okapi/api/units/RQuantity.hpp
Octogonapus: AFAIK this seems right. @Breq16 try removing these methods and compiling some code that used them.

theol0403:
> I agree they are useful and I don't think it would be difficult to implement. Wouldn't round, floor, and ceil simply return the same dimensions as they accepted?

The problem is that the rounding depends on the internal unit - round would work differently with degrees than radians for example, so we would need to specify what unit we want to round to.

Octogonapus:
> The problem is that the rounding depends on the internal unit - round would work differently with degrees than radians for example, so we would need to specify what unit we want to round to.

Oh, right. I forgot about that. I think it makes sense that we'd require the user to supply a specific unit to round in. For example, `floor<inch>(1.2_in) == 1_in`, `round<meter>(47_in) == 1_m`.

include/okapi/api/units/RQuantity.hpp
breqdev: That seems to work! I'm kind of a noob when it comes to C++, so thanks for letting me know about this behavior.

theol0403:
Finally, I think it would be a good idea to write some simple tests for the functionality of these new methods. Even if the highest chance of something breaking is within compile-time, there is always a possibility of having issues with certain types, or silly things like swapped parameters. Testing is definitely necessary for the rounding methods if you choose to implement them.

Writing tests are fairly straightforward with googletest, but if you aren't comfortable with the testing system I'm happy to write them.

breqdev:
I had a working implementation of `pow()` that used a `std::ratio` as a template parameter (so cubing a number would be e.g. `pow<std::ratio<3>>(3_in)`). The syntax isn't ideal (which is why I didn't include it in the initial PR), but I can't think of a better way to implement it. I'd be happy to add this as well as some more user-friendly `square()` and `cube()` functions.

Correct me if I'm wrong here, but it seems like the syntax for the rounding would be `floor(1.2_in, inch)`, not `floor<inch>(1.2_in)`, because `inch` is an instance of `QLength`, not a type? If so, implementing the rounding functions this way shouldn't be a problem.

It seems like the other log and exponent functions wouldn't really work because of the way units are handled (for instance, `log(1_m)` and `log(1_ft)` can't both return zero).

I don't really have any experience writing tests, but I could give it a try. (It seems like the correct place to add them is in `test/unitTests.cpp`?) I'll read the googletest documentation and try to write some.

theol0403:
For `pow`, you should be able to get away with accepting an `int` as the template parameter (it's one of the few things that templates allow), and doing `pow<3>(3_in)`. 

Humm, you are correct. I'm not actually sure how you would go about this then - the unit parameter would be an RQuantity of the same dimension as the value parameter, but then the only thing that you can use to tell the units apart is the ratio to their underlying unit (`inch.getValue() == 0.0254`). Maybe it would work if you divided the value by the unit ratio, used the rounding function, and then multiplied the value by the ratio. 

Yep, that's where you should put the unit tests. You just make a test case, and the testing framework will automagically add it to the tests which the CI will run. You can also run the tests locally by running the cmake commands outlined by the azure build file (in the project root).

breqdev:
Cool, I'll get to work on the unit tests.

With regards to `pow`, it seems like it'd be best to allow users to raise an RQuantity to a fractional exponent as well (such as `pow<std::ratio<1, 4>>` for a fourth root). It seems like it's possible to have both this version and the nicer int version. Is it worth keeping both, or just the int version?

theol0403:
Yeah, I'd keep them both as overloads. It might also be nice to have a `root<int>` method to compliment the `pow<int>` method, and then the `pow<R>` is to do both.

include/okapi/api/units/RQuantity.hpp
```diff
@@ -213,6 +213,30 @@ pow(const RQuantity<M, L, T, A> &lhs) {
...
+template <int R, typename M, typename L, typename T, typename A>
+constexpr RQuantity<std::ratio_divide<M, std::ratio<R>>,
+                    std::ratio_divide<L, std::ratio<R>>,
+                    std::ratio_divide<T, std::ratio<R>>,
+                    std::ratio_divide<A, std::ratio<R>>>
+root(const RQuantity<M, L, T, A> &lhs) {
+  return RQuantity<std::ratio_divide<M, std::ratio<R>>,
+                   std::ratio_divide<L, std::ratio<R>>,
+                   std::ratio_divide<T, std::ratio<R>>,
+                   std::ratio_divide<A, std::ratio<R>>>(std::pow(lhs.getValue(), 1/R));
```
theol0403: You will run into integer arithmetic problems here. Either use `1.0` or there might be a root function in the standard lib.

test/unitTests.cpp
```diff
@@ -44,3 +46,82 @@ TEST(UnitTests, UnaryMinusTest) {
...
   EXPECT_DOUBLE_EQ(std::sqrt(2), (2 * meter2).sqrt().convert(meter));
 }
+
+TEST(UnitTests, AbsFreeFunction) {
+  EXPECT_EQ(abs(3.0_m), 3.0_m);
+  EXPECT_EQ(abs(-3.0_m), 3.0_m);
+}
+
+TEST(UnitTests, PowRootTest) {
+  EXPECT_DOUBLE_EQ((pow<2>(3_in)).convert(inch2), 9.0);
```
Octogonapus, line 56: I can't convince myself that this test passes. Wouldn't this call `std::pow(0.0762, 2)` which is definitely not 9 inches?

breqdev: The call to `.convert(inch2)` converts the result of pow (0.00581 m²) to square inches (9.0 in²)

PR #447, opened 2020-07-13
Enable autobrief

theol0403:
### Description of the Change

Enables doxygen autobrief which uses the first sentence of the function comment as the `\brief` command.

Before: 
image
image

After:
image
image

### Motivation

Makes the docs look better and easier navigation.

### Possible Drawbacks

I have not looked through all pages to see how this changes things, though I think this change is a very minor risk.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -195,7 +195,7 @@ SHORT_NAMES            = NO
 # description.)
 # The default value is: NO.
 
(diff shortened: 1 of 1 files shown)


Octogonapus (approved):
I would appreciate it if you could spot-check various docs pages, but otherwise, this looks good.

theol0403:
Yeah I looked through some pages and it looks fine.

PR #452, opened 2020-08-03
#444: Add V5 IMU Wrapper Class

### LINKED ISSUE #444: Add wrapper class for V5 IMU device similar to the adiGyro class
Would like to be able to use the new V5 inertia measuring unit in much the same way that the old ADI Gyro was

Similar class to the adiGyro. Perhaps with a flag that would allow one to change the X,Y,Z (yaw/pitch/roll) value returned.

Using the raw PROS IMU class

**Versions**
kernel 3.2.1 pros-mainline
okapilib 4.0.3 pros-mainline

**Additional context**
N/A

invalidflaw:
### Description of the Change

Adds a new wrapper class derived from ContinuousRotarySensor for the V5 IMU. The constructor accepts a V5 port number and an enum axis definition to measure. The z(yaw) axis is configured as default. All axes return an angle in the range of -180 to 180, much like the existing ADIgyro class does. To measure more axes, separate IMU instances should be made by the user.

As suggested in the issue, a value, "offset" is used to reset the sensor. The PROS API does not allow for an easy zero process, so the `IMU::reset()` command will change the offset parameter to match the current measured angle. In the event of an `IMU::get()` command occurs outside the range of -180 to 180, the offset is inverted by adding -360*(offset/sign(offset)) to the original offset. 

Because the reset command does not calibrate the IMU, I added an `IMU::calibrate()` method. The constructor does not call this as the v5 port is automatically configured and is calibrated on brain startup. Any further calibration should be controlled by the user.

Because the IMU is an accelerometer as well as a gyro, I added an `IMU::getAccl()` command to retreive the acceleration value along the instanced axis.

### Motivation

Adding a wrapper class for the IMU will allow it to be integrated into odometry and chassis models at a later date. Implementing the sensor within okapi allows users to program their robot using the same namespace.

### Possible Drawbacks

The V5 IMU chooses one of 6 orientations based on the calibration process, then assigns the XYZ axes accordingly. I am unsure how/if this should be documented.

### Verification Process

I placed the IMU on a robot chassis read the gyro data using the PROSC API and the new IMU class. Both classes returned the same euler angles.

To test the reset function, I set the IMU to measure robot yaw. I started the program, and rotated the chassis 90 degrees, after which I pressed a button on my controller which was set to run the reset function. After doing so, the robot continued to read -180 to 180 with the 0 now located at the original 90 degrees. 

### Applicable Issues

Closes #444

### DIFF
--- a/include/okapi/impl/device/rotarysensor/IMU.hpp
+++ b/include/okapi/impl/device/rotarysensor/IMU.hpp
@@ -34,15 +34,17 @@ class IMU : public ContinuousRotarySensor {
   /**
    * Get the current rotation about `iaxis`.
    *
-   * @return The current sensor value or one of the following values of `errno`:
+   * This may set the following values of `errno` on error:
    * - `ENXIO` - The given `iport` is not in range of the V5 ports (1-21).
    * - `ENODEV` - The port cannot be configured as an IMU.
    * - `EAGAIN` - The sensor is calibrating.
+   *
+   * @return The current sensor value or `PROS_ERR`.
    */
   double get() const override;
 
   /**
-   * Get the current sensor value remapped into the target range (`[1800, -1800]` by default).
+   * Get the current sensor value remapped into the target range (`[-1800, 1800]` by default).
    *
    * @param iupperBound The upper bound of the range.
    * @param ilowerBound The lower bound of the range.
@@ -53,10 +55,12 @@ class IMU : public ContinuousRotarySensor {
   /**
    * Get the current acceleration along `iaxis`.
    *
-   * @return The current sensor value or one of the following values of `errno`:
+   * This may set the following values of `errno` on error:
    * - `ENXIO` - The given `iport` is not in range of the V5 ports (1-21).
    * - `ENODEV` - The port cannot be configured as an IMU.
    * - `EAGAIN` - The sensor is calibrating.
+   *
+   * @return The current sensor value or `PROS_ERR`.
    */
   double getAcceleration();
 
@@ -71,10 +75,12 @@ class IMU : public ContinuousRotarySensor {
    * Calibrate the IMU. Resets the rotation value to zero. Calibration is expected to take two
    * seconds, but is bounded to five seconds.
    *
-   * @return `1` on success, or one of the following values of `errno`:
+   * This may set the following values of `errno` on error:
    * - `ENXIO` - The given `iport` is not in range of the V5 ports (1-21).
    * - `ENODEV` - The port cannot be configured as an IMU.
-   * - `EAGAIN` - The sensor is calibrating or timed out calibrating.
+   * - `EAGAIN` - The sensor is calibrating.
+   *
+   * @return `1` or `PROS_ERR`.
    */
   std::int32_t calibrate();
 
@@ -82,10 +88,12 @@ class IMU : public ContinuousRotarySensor {
    * Get the sensor value for use in a control loop. This method might be automatically called in
    * another thread by the controller.
    *
-   * @return the current sensor value, or one of the following values of `errno`
-   * - `ENXIO` - The given iport is not in range of the V5 ports (1-21).
+   * This may set the following values of `errno` on error:
+   * - `ENXIO` - The given `iport` is not in range of the V5 ports (1-21).
    * - `ENODEV` - The port cannot be configured as an IMU.
    * - `EAGAIN` - The sensor is calibrating.
+   *
+   * @return The current sensor value or `PROS_ERR`.
    */
   double controllerGet() override;
 
@@ -96,10 +104,12 @@ class IMU : public ContinuousRotarySensor {
   /**
    * Get the current rotation about `iaxis` from the IMU. The internal offset is not accounted for.
    *
-   *@return the current sensor value or one of the following values of `errno`
-   * - `ENXIO` - The given iport is not in range of the V5 ports (1-21).
+   * This may set the following values of `errno` on error:
+   * - `ENXIO` - The given `iport` is not in range of the V5 ports (1-21).
    * - `ENODEV` - The port cannot be configured as an IMU.
    * - `EAGAIN` - The sensor is calibrating.
+   *
+   * @return The current sensor value or `PROS_ERR`.
    */
   double readAngle() const;
 
--- a/include/okapi/impl/device/rotarysensor/IMU.hpp
+++ b/include/okapi/impl/device/rotarysensor/IMU.hpp
@@ -12,88 +12,87 @@
 #include "okapi/api/device/rotarysensor/continuousRotarySensor.hpp"
 
 namespace okapi {
-enum class IMUAxes{
+enum class IMUAxes {
   z, ///< Yaw Axis
   y, ///< Pitch Axis
   x  ///< Roll Axis
 };
 
-class IMU : public ContinuousRotarySensor{
+class IMU : public ContinuousRotarySensor {
   public:
   /**
-   * An inertial sensor on the given port. If the port has not previously been configured as an IMU,
-   * then the constructor will block for 2 seconds for calibration. The IMU returns double angle results
+   * An inertial sensor on the given port.  The IMU returns an angle
+   * about the selected axis in degrees.
    *
    * @param iport: The port to use the inertial sensor from
    * @param iaxis: The axis of the inertial sensor to measure, default z
    */
-   IMU(std::uint8_t iport,IMUAxes iaxis = IMUAxes::z);
+  IMU(std::uint8_t iport, IMUAxes iaxis = IMUAxes::z);
 
-   virtual ~IMU();
+  virtual ~IMU();
 
-   /**
-    * Get the current rotation about iaxis
-    *
-    *@return the current sensor value or one of the following values of `errno`
-    * `ENXIO` - the given iport is not in range of the V5 ports (1-21)
-    * `ENODEV` - the port cannot be configured as an IMU
-    * `EAGAIN` - the sensor is calibrating
-    */
-    double get() const override;
+  /**
+   * Get the current rotation about iaxis.
+   *
+   *@return the current sensor value or one of the following values of `errno`
+   * `ENXIO` - the given iport is not in range of the V5 ports (1-21)
+   * `ENODEV` - the port cannot be configured as an IMU
+   * `EAGAIN` - the sensor is calibrating
+   */
+  double get() const override;
 
-    /**
-     * Get the current sensor value remapped into the target range ([1800, -1800] by default).
-     *
-     * @param iupperBound the upper bound of the range.
-     * @param ilowerBound the lower bound of the range.
-     * @return the remapped sensor value.
-     */
-    double getRemapped(double iupperBound = 180, double ilowerBound = -180) const
-      __attribute__((optimize(3)));
+  /**
+   * Get the current sensor value remapped into the target range ([1800, -1800] by default).
+   *
+   * @param iupperBound the upper bound of the range.
+   * @param ilowerBound the lower bound of the range.
+   * @return the remapped sensor value.
+   */
+  double getRemapped(double iupperBound = 1800, double ilowerBound = -1800) const;
 
-      /**
-       * Get the current acceleration along iaxis
-       *
-       *@return the current sensor value or one of the following values of `errno`
-       * `ENXIO` - the given iport is not in range of the V5 ports (1-21)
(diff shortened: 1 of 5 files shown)


include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -0,0 +1,99 @@
+/*
```
Octogonapus, line 1: Please run clang-format (there is a script in the project root) and format documentation as full sentences.

include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Get the current sensor value remapped into the target range ([1800, -1800] by default).
+     *
+     * @param iupperBound the upper bound of the range.
+     * @param ilowerBound the lower bound of the range.
+     * @return the remapped sensor value.
+     */
+    double getRemapped(double iupperBound = 180, double ilowerBound = -180) const
+      __attribute__((optimize(3)));
```
Octogonapus: You can remove this attribute.

include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+      /**
+       * Get the current acceleration along iaxis
+       *
+       *@return the current sensor value or one of the following values of `errno`
+       * `ENXIO` - the given iport is not in range of the V5 ports (1-21)
+       * `ENODEV` - the port cannot be configured as an IMU
+       * `EAGAIN` - the sensor is calibrating
+       */
+    double getAcc();
```
Octogonapus: I would rename this to `getAcceleration`

src/impl/device/rotarysensor/IMU.cpp
```diff
@@ -0,0 +1,77 @@
...
+    case IMUAxes::z:
+    return accel.z;
+    break;
+  }
+  return -1;
+}
+
+std::int32_t IMU::reset() {
+  const auto value = get() - offset;
+  offset = -1 * value;
```
Octogonapus: This is a weird way to do the offset. I would rather you did `offset = get()`.

include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Reset the sensor to zero.
+     *
+     * @returns `1` on success
+     */
+    std::int32_t reset() override;
```
Octogonapus: Return `void` instead.

Octogonapus (changes requested):

src/impl/device/rotarysensor/IMU.cpp
invalidflaw: In order to assign a new offset, the old offset must be removed. To implement this, I will first zero offset before running ```offset = get();```.

include/okapi/impl/device/rotarysensor/IMU.hpp
invalidflaw: This poses an issue. The ContinuousRotarySensor class requires this function to be of the form ```virtual std::int32_t reset() = 0;```. As such, this must return a std::int32_t. I propose that this return a 1 as this identical to other ContinuousRotarySensor reset success behavior.

include/okapi/impl/device/rotarysensor/IMU.hpp
invalidflaw: I have made this change, it will be present in the next commit

src/impl/device/rotarysensor/IMU.cpp
Octogonapus: Yeah, you are right; what I meant was more like `offset = <get the axis value from PROS>` instead of `offset = get()`.

include/okapi/impl/device/rotarysensor/IMU.hpp
Octogonapus: Right, I forgot. `std::int32_t` is fine.

src/impl/device/rotarysensor/IMU.cpp
invalidflaw: That can be done. Take a look at my last commit. Would you like me to use the pros get in a switch or keep it as-is?

Octogonapus: I think it would be good to pull this out into a method
```
  pros::c::euler_s_t eu = pros::c::imu_get_euler(port);
  static double angle = 0;
  switch (axis) {
  case IMUAxes::x:
    angle = eu.roll;
    break;
  case IMUAxes::y:
    angle = eu.pitch;
    break;
  case IMUAxes::z:
    angle = eu.yaw;
    break;
  }
```
and use it to get the offset. Also, you don't need to have `angle` be `static`.

invalidflaw: I have added a private `readAngle()` method, and implemented it in `get()` and `calibrate()`. The commit compiles, however I want to test it on my robot later tonight.

invalidflaw: Just tested my latest pull request, the new private member function `double readAngle()` works as expected.

Octogonapus (dismissed):
This is approved pending CI passes and a kernel bug fix.

include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * An inertial sensor on the given port. The IMU returns an angle about the selected axis in
+   * degrees.
+   *
+   * @param iport The port to use the inertial sensor from.
+   * @param iaxis The axis of the inertial sensor to measure, default `IMUAxes::z`.
+   */
+  IMU(std::uint8_t iport, IMUAxes iaxis = IMUAxes::z);
+
+  virtual ~IMU();
```
theol0403: Explicitly defining a virtual destructor isn't needed as it inherits the virtual destructor from base classes.

include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -0,0 +1,124 @@
...
+   * Get the current acceleration along `iaxis`.
+   *
+   * This may set the following values of `errno` on error:
+   * - `ENXIO` - The given `iport` is not in range of the V5 ports (1-21).
+   * - `ENODEV` - The port cannot be configured as an IMU.
+   * - `EAGAIN` - The sensor is calibrating.
+   *
+   * @return The current sensor value or `PROS_ERR`.
+   */
+  double getAcceleration();
```
theol0403: Can this function be const?

src/impl/device/rotarysensor/IMU.cpp
```diff
@@ -0,0 +1,116 @@
...
+IMU::~IMU() = default;
+
+double IMU::get() const {
+  double angle = readAngle() - offset;
+
+  if (angle == PROS_ERR) {
+    return PROS_ERR;
+  }
+
+  if (angle > 180 || angle < -180) {
```
theol0403: This functionality matches the recently added okapilib helper functions to constrain an angle to 180 degrees. Perhaps use those instead?

theol0403 (dismissed):
Just some tiny nits. Otherwise looks good.

> All axes return an angle in the range of -180 to 180

Imo I don't think this is nice to use. You have to deal with the input to your control algorithms jumping if you don't constrain the angle after you do any `setpoint-reading` calculations. The IMU can return an unconstrained range through a different API method, which might be nicer to use. Up to you two, it's not a huge deal either way.

invalidflaw:
When I was developing this class, I settled on the [-180, 180] range for a couple of reasons. 

First, the PROS API for the IMU is pretty limited. I don't believe there is a command which gives me an unbounded rotation for all 3 axes. In my testing, `pros::c::get_yaw()`, `pros::c::get_pitch()` and `pros::c::get_roll()` all simply return the euler angles, which I am using in this class. 

The second reason I chose this range was to match the existing gyro better. I would need to test the ADIgyro again, but I believe its range is [-1800,1800]. In order to retain some compatibility, I wanted these to use the same general tracking behavior. It is possible to implement unbounded counting in software, but I believe we should then update ADIgyro as well to reflect this behavior change. Consistent behavior will be needed for my next project of adding a gyro to odometry.

include/okapi/impl/device/rotarysensor/IMU.hpp
Octogonapus: ADIGyro probably shouldn't have one, either.

theol0403:
Sounds good. Did you want to use https://github.com/OkapiLib/OkapiLib/blob/a9b3eadb52d8e1d46b1f619699ebdb5732dbba0f/include/okapi/api/odometry/odomMath.hpp#L63 for the angle wrapping? I'm sure you could also move the function to `mathUtil.hpp` which puts the function in a centrally available place. Doing this has the added benefit of being able to remap values which require more than 360 degrees of correction.

Octogonapus:
I agree with Theo, I think moving that function to `mathUtil.hpp` and then using it here would be a good idea.

invalidflaw:
I'm a little confused by this. Is the proposal to make the IMU and ADIgyro class return unbounded rotation values, and then convert that using `constrainAngle180`?

Octogonapus:
> I'm a little confused by this. Is the proposal to make the IMU and ADIgyro class return unbounded rotation values, and then convert that using `constrainAngle180`?

I believe the proposal is to use `constrainAngle180` [here](https://github.com/OkapiLib/OkapiLib/blob/e86428962621e9d0a0017c8e0efd7500fb4126c3/include/okapi/api/odometry/odomMath.hpp#L63) instead of copying that functionality here.

invalidflaw:
In that case, the next step is to remove `getRemapped` from both IMU and ADIGyro classes. Correct?

Octogonapus:
> In that case, the next step is to remove `getRemapped` from both IMU and ADIGyro classes. Correct?

No. We can't break backward compatibility.

invalidflaw:
I must first apologize. I think I am more confused than before. Looking at differences between ADIGyro and IMU, IMU `get` returns a bounded -180 to 180 degree angle value, while the ADIGyro returns unbounded. I see value in making these two `get` functions return the same values. As to the `getRemapped` issue, is the goal to replace the mathUtil `remapRange` function with `constrainAngle180`? My only concern for that is you lose the ability to specify the return range. Is that considered acceptable for most uses?

theol0403:
Sorry for any confusion, all I was saying was to replace
```cpp
  if (angle > 180 || angle < -180) {
    angle += std::copysign(360, offset);
  }
```
with 
```cpp
angle = constrainAngle180(angle);
```
to remove code duplication.

invalidflaw:
Ok, that makes sense. As to the lack of angle wrapping, should I spend a bit of time and make the IMU return unbounded by default? The existing ADIgyro returns unbounded. I'm a little worried that if these two don't match, our upcoming project to implement gyro based odometry may become difficult.

Octogonapus:
> 
> 
> Ok, that makes sense. As to the lack of angle wrapping, should I spend a bit of time and make the IMU return unbounded by default? The existing ADIgyro returns unbounded. I'm a little worried that if these two don't match, our upcoming project to implement gyro based odometry may become difficult.

I still am against supporting (by default) odometry incorporating the ADI gyro. I don't think it's worth you putting in the extra work here to match that functionality. We'd probably want to convert back to a bounded angle anyway.

invalidflaw:
Implemented the suggestion from @theol0403 . I would like to give it a test, but if it works and there are no further concerns. I think this is ready for review.

theol0403:
Looks good. Ideally the `constrainAngle180` method should be moved to `mathUtil.hpp`, so it can be in a common place (IMU should not depend on OdomMath). However, if @Octogonapus is happy, I'm happy.

Octogonapus:
I am fine with leaving it in odomMath.

I pushed an update to fix the return value from `calibrate` in the event that it times out. We will need to test this function once a new kernel release is out that contains https://github.com/purduesigbots/pros/pull/236. So I am essentially approving this; we just need a kernel release now.

invalidflaw:
Just gave the latest code a test, works as expected.

Octogonapus:
@invalidflaw or @kunwarsahni01 can one of you please test the IMU, especially calibration?

invalidflaw:
I'll give it a try tomorrow morning.

invalidflaw:
Just tested this morning. The IMU works as expected. Calibration and IMU disconnect are handled properly.

PR #455, opened 2020-10-30
Support new distance, optical, and rotation sensors

dcieslak19973:
### Description of the Change

PROS now supports the new distance, optical, and rotation sensors

### Motivation

Additionally, allow the chassis to support odometry with the new rotation sensor

### Possible Drawbacks

N/A

### Verification Process

Passed physical testing with the distance and optical sensors. Have not physically tested the rotation sensors

### Applicable Issues

N/A


Octogonapus:
Thanks for the quick PR! After your update your branch with `develop`, please use the kernel's C API instead of its C++ API when implementing OkapiLib classes and try to use full sentences in the documentation. Then I will take another review pass.

theol0403:
Thanks for making this PR!

Something went wrong when you tried to update with develop. The diff should only be showing the changes that you wrote.

- This PR needs to target `develop`, not `master`. You should be able to change that. Merging develop into your PR was not the right thing to do.
- The PR should ideally not be adding merge commits. You should instead rebase your commits onto develop, minus the merge commit. I think the easiest approach would be to just start over and then force-push your branch. Your commits conflict with develop, as you updated the kernel in this PR, so you need to get rid of that - but since that commit contains some of your work, you can't just drop the commit in an interactive rebase (which otherwise would have been the solution).

dcieslak19973:
Will resubmit

PR #458, opened 2020-10-31
Add optical distance rotation sensors

dcieslak19973:
### Description of the Change

Add support for new Color, Distance, and Rotation sensors.

### Motivation

Allow Chassis Builder to use the new RotationSensor

Provide wrapper classes for Color and Distance sensors.

The Rotation sensor is modeled after the ADI Encoder class

The Distance sensor is modeled after the ultrasonic sensor class

### Possible Drawbacks

N/A

### Verification Process

Verified physically with the color and distance sensors. Have not physically tested the rotation sensor.

### Applicable Issues

N/A


dcieslak19973:
master not dev...

PR #459, opened 2020-10-31
Adding color, distance, rotation sensors

dcieslak19973:
Allows chassis builder to reference new RotationSensor

### Description of the Change

Add wrappers for newly supported PROS API for the Color, Distance, and Rotation sensors

### Motivation

Allow Chassis to reference the new Rotation Sensor for odometry

### Possible Drawbacks

N/A

### Verification Process

Tested Color and Distance sensors physically

### Applicable Issues

N/A


include/okapi/impl/device/distanceSensor.hpp
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * This function uses the following values of errno when an error state is
+   * reached:
+   * ENXIO - The given value is not within the range of V5 ports (1-21).
+   * ENODEV - The port cannot be configured as an Distance Sensor
+   *
+   * \return The confidence value or PROS_ERR if the operation failed, setting
+   * errno.
+   */
+  virtual std::int32_t get_confidence();
```
theol0403: Methods that are specific to the device, and that do not override a base method, don't need to be virtual.

theol0403: Methods that are `const` should be marked as such.

include/okapi/impl/device/distanceSensor.hpp
```diff
@@ -0,0 +1,96 @@
...
+   * This is a value that has a range of 0 to 63. 63 means high confidence,
+   * lower values imply less confidence. Confidence is only available
+   * when distance is > 200mm.
+   *
+   * This function uses the following values of errno when an error state is
+   * reached:
+   * ENXIO - The given value is not within the range of V5 ports (1-21).
+   * ENODEV - The port cannot be configured as an Distance Sensor
+   *
+   * \return The confidence value or PROS_ERR if the operation failed, setting
```
theol0403: I believe the standard in this repo is to prefix doxygen commands with `@`.

include/okapi/impl/device/distanceSensor.hpp
```diff
@@ -0,0 +1,96 @@
...
+  virtual double controllerGet() override;
+
+  /**
+   * Get the confidence in the distance reading
+   *
+   * This is a value that has a range of 0 to 63. 63 means high confidence,
+   * lower values imply less confidence. Confidence is only available
+   * when distance is > 200mm.
+   *
+   * This function uses the following values of errno when an error state is
```
theol0403: Is this errno documentation supposed to be here? I don't see it in any of the other device headers.

include/okapi/impl/device/distanceSensor.hpp
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Get the sensor value for use in a control loop. This method might be automatically called in
+   * another thread by the controller. Calls get().
+   */
+  virtual double controllerGet() override;
+
+  /**
+   * Get the confidence in the distance reading
+   *
+   * This is a value that has a range of 0 to 63. 63 means high confidence,
```
theol0403: I'm not sure if this formatting comes from the PROS repo, but the comments can be condensed while still respecting column limit. To do this, just combine the lines together, and then let clang-format split them up properly. For example, on this comment I got
```cpp
   * This is a value that has a range of 0 to 63. 63 means high confidence, lower values imply less
   * confidence. Confidence is only available when distance is > 200mm.
   *
   * This function uses the following values of errno when an error state is reached:
```

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   * ```
+   *
+   * @param iport The V5 port the device uses
+   */
+  OpticalSensor(std::uint8_t iport);
+
+  virtual ~OpticalSensor();
+
+  /**
+   * Returns the current sensor's hue value (0-359.99).
```
theol0403: To better match okapi style, you might want to do:
```suggestion
   * Returns The current sensor's hue value in the range [0, 360).
```
Also, you could surround the range notation with double backticks, if it makes the documentation look nicer.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   *
+   * @return current value
+   */
+  virtual double get();
+
+  /**
+   * Returns the current sensor's brightness value (0-1.0).
```
theol0403:
```suggestion
   * Get the current sensor's brightness value in the range [0, 1].
```

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   *
+   * @return current value
+   */
+  virtual double get_brightness();
+
+  /**
+   * Returns the current sensor's saturation value (0-1.0).
```
theol0403:
```suggestion
   * Get the current sensor's saturation value in the range [0, 1].
```

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   * @return current value
+   */
+  virtual double get_saturation();
+
+  /**
+   * Returns the current sensor's LED brightness
+   *
+   * value that ranges from 0 to 100
+   *
+   * @return current value
```
theol0403:
```suggestion
   * @return The current sensor value in the range [0, 100].
```
Then you can delete the middle two lines of the comment.

Same with below.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   */
+  virtual int32_t get_led_brightness();
+
+  /**
+   * Set the pwm value of the White LED on the sensor
+   *
+   * value that ranges from 0 to 100
```
theol0403: It's important that the doxygen `@param` exists for every parameter, which doxygen needs to properly generate documentation.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   */
+  virtual int32_t set_led_brightness(int32_t v);
+
+  /**
+   * This is not available if gestures are being detected. proximity has
+   * a range of 0 to 255.
```
theol0403: This needs a title. 
Like I've suggested above, the range information can be included with the `@return` command.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   * @return current value
+   */
+  virtual int32_t get_led_brightness();
+
+  /**
+   * Set the pwm value of the White LED on the sensor
+   *
+   * value that ranges from 0 to 100
+   */
+  virtual int32_t set_led_brightness(int32_t v);
```
theol0403: As per okapi convention, all the parameters need the `i` prefix in their name. What is `v`? Should it be `ipwm`?
Maybe the function should be renamed to `set_led_pwm`  to stay consistent with `get_led_pwm` and the PROS API.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,118 @@
...
+   *
+   * This function uses the following values of errno when an error state is
+   * reached:
+   * ENXIO - The given value is not within the range of V5 ports (1-21).
+   * ENODEV - The port cannot be configured as an Optical Sensor
+   *
+   * \return rgb value if the operation was successful or an optical_rgb_s_t
+   * with all fields set to PROS_ERR if the operation failed, setting errno.
+   */
+  virtual pros::c::optical_rgb_s_t get_rgb();
```
theol0403: Good example to summarise some of the changes I'm requesting:
- no need to be virtual
- should be const
- needs to be camel case
- comments probably should not have errno documentation
- doxygen `@` instead of `\`
- in the documentation, surround the `optical_rgb_s_t` with backticks so that doxygen can automatically hyperlink it.
- line wrap can be reduced
- period at the end of sentences, especially the title

include/okapi/impl/device/rotarysensor/rotation.hpp
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @param iport The V5 port the device is on
+   * @param ireversed Whether the sensor is reversed.
+   */
+  RotationSensor(std::uint8_t iport, bool ireversed = false);
+
+  /**
+   * Get the current sensor value.
+   *
+   * @return the current sensor value, or `PROS_ERR` on a failure.
```
theol0403: You can probably find-replace all instances of this.
```suggestion
   * @return The current sensor value or ``PROS_ERR``.
```

Also find-replace `@return the` to `@return The`

src/impl/device/distanceSensor.cpp
```diff
@@ -0,0 +1,36 @@
...
+ */
+#include "okapi/impl/device/distanceSensor.hpp"
+
+namespace okapi {
+DistanceSensor::DistanceSensor(std::uint8_t iport, std::unique_ptr<Filter> ifilter)
+  : filter(std::move(ifilter)) {
+  port = iport;
+}
+
+DistanceSensor::~DistanceSensor() = default;
```
theol0403: The destructor, if it's default, should not be here and just marked as default in the header.

src/impl/device/rotarysensor/rotation.cpp
```diff
@@ -0,0 +1,31 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+#include "okapi/impl/device/rotarysensor/rotation.hpp"
+
+namespace okapi {
+RotationSensor::RotationSensor(const std::uint8_t iport, const bool ireversed) {
+  port = iport;
+  pros::c::rotation_set_reversed(port, ireversed);
```
theol0403: Up to @Octogonapus , but maybe `reversed` should be an internal flag like with `Motor`, so that multiple instances of the class won't be interfering with the direction of the sensor.

src/impl/device/rotarysensor/rotation.cpp
```diff
@@ -0,0 +1,31 @@
...
+}
+
+double RotationSensor::get() const {
+  const double angle = pros::c::rotation_get_position(port);
```
theol0403: You can probably just directly return the value, not sure if there is some standard being met here.

theol0403 (changes requested):
Thanks, and thanks again for making the PR! Though you could have just changed the target of the last PR, and then force-pushed your branch, which would have fixed the PR.

You need to change all the method names to camelCase to conform to okapi convention, I know it's a bit of a pain. I have it set up in vim to switch the case on command, so if you want I could do this for you.

I know a lot of my comment suggestions are not your fault, you just copied the PROS comments, but I think they should be fixed for okapi.

Thanks!

CONTRIBUTING.md
```diff
@@ -95,6 +95,23 @@ cd OkapiLib
 git submodule update --init --recursive
 ```
 
+To build the Okapi library:
```
theol0403: I just realized you wrote this as well. 

The `mkdir template` command should not be needed. 
>  Which will result in the okapilib.a file being created in the `bin` directory.

Yes, but what you care about is that it will result in `okapilib@version.zip` created in the root directory, which you can then import into the PROS conductor and then apply it to your projects.

Also, might want to add that the commands need to be run on linux, with gcc, make, cmake, and git installed.

>  To build the Okapi library:

To build OkapiLib:

include/okapi/impl/device/distanceSensor.hpp
dcieslak19973: Addressed in most recent commit

dcieslak19973: Addressed in most recent commit

include/okapi/impl/device/distanceSensor.hpp
dcieslak19973: Addressed in most recent commit

include/okapi/impl/device/distanceSensor.hpp
dcieslak19973: Attempted; clang-format not working for me locally at present though

dcieslak19973:
Bump - waiting on anything from me?

src/impl/device/opticalSensor.cpp
```diff
@@ -0,0 +1,55 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+#include "okapi/impl/device/opticalSensor.hpp"
+
+namespace okapi {
+OpticalSensor::OpticalSensor(std::uint8_t iport) {
+  port = iport;
+  pros::c::optical_disable_gesture(port);
```
Octogonapus: Why do you disable gestures?

dcieslak19973: My understanding is the device is either in color-detection mode or gesture mode. It cannot do both at the same time, at least based on how I read the V5 code documentation. Figured defaulting to color-detection mode was a reasonable choice.

include/okapi/impl/device/distanceSensor.hpp
```diff
@@ -0,0 +1,75 @@
...
+   * @param iport The V5 port the device uses.
+   * @param ifilter The filter to use for filtering the distance measurements.
+   */
+  DistanceSensor(std::uint8_t iport,
+                 std::unique_ptr<Filter> ifilter = std::make_unique<PassthroughFilter>());
+
+  virtual ~DistanceSensor();
+
+  /**
+   * @return The current filtered sensor value in mm.
```
theol0403, line 35: This needs a title to be parsed by doxygen.

include/okapi/impl/device/distanceSensor.hpp
```diff
@@ -0,0 +1,75 @@
...
+   * This is a value that has a range of ``[0, 400]``. A 18" x 30" grey card will return a value
+   * of approximately ``75`` in typical room lighting.
+   *
+   * @return The size value in the range ``[0, 400]`` or ``PROS_ERR`` if the operation failed,
+   * setting errno.
+   */
+  std::int32_t getObjectSize() const;
+
+  /**
+   * @return The object velocity in m/s.
```
theol0403, line 68: Needs title.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,122 @@
...
+   * @param iport The V5 port the device uses.
+   * @param ioutput Which sensor output to return from (@ref okapi::OpticalSensor::get).
+   * @param ifilter The filter to use to filter the sensor output. Only the selected output (via
+   * ``ioutput``) is filtered; the other outputs are untouched.
+   */
+  OpticalSensor(std::uint8_t iport,
+                OpticalSensorOutput ioutput = OpticalSensorOutput::hue,
+                std::unique_ptr<Filter> ifilter = std::make_unique<PassthroughFilter>());
+
+  virtual ~OpticalSensor();
```
theol0403: Should be defaulted here instead of in the cpp file.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * @return The current filtered value of the selected output (configured by the constructor).
```
theol0403, line 48: All these below need a title for doxygen to parse.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,122 @@
...
+  int32_t getLedPWM();
+
+  /**
+   * Set the PWM value of the white LED.
+   *
+   * @param value The PWM value in the range ``[0, 100]``.
+   * @return ``1`` if the operation was successful or ``PROS_ERR`` if the operation failed, setting
+   * ``errno``.
+   */
+  int32_t setLedPWM(uint8_t value);
```
theol0403: I think we need to stick to the `ivalue` pattern.

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * This is not available if gestures are being detected.
```
theol0403: While this does have a title, it ideally needs to be a little blurb about the method, because the title shows up in the documentation overview like 
image

include/okapi/impl/device/opticalSensor.hpp
```diff
@@ -0,0 +1,122 @@
...
+   *
+   * @return The current proximity value in the range ``[0, 255]``.
+   */
+  int32_t getProximity();
```
theol0403: All methods that have `get` in their name should be const.

theol0403 (changes requested):
Thanks for fixing most of the issues!
I'll take a bigger look (in a few days), I'll just leave some comments now.

I've noticed that by making the comments more consise, some have lost the title. Doxygen needs a title to parse properly. Duplication is okay, for example:
image

include/okapi/impl/device/distanceSensor.hpp
Octogonapus: Only for the overview. Do you really care about that?

include/okapi/impl/device/opticalSensor.hpp
Octogonapus: Why?

theol0403: Because they just have `DistanceSensor::~DistanceSensor() = default;` in the source file.

theol0403: Oh oops I confused the class. Maybe the optical sensor has a non default destructor, in that case it's fine.

theol0403 (dismissed):
Everything looks good! However, I would still recommend implementing a user-side flag for the direction of the rotation sensor, like with `okapi::Motor`. 

https://github.com/OkapiLib/OkapiLib/blob/e1c8aece8691222e675233ddfade5485db4ea19b/include/okapi/impl/device/motor/motor.hpp#L455
https://github.com/OkapiLib/OkapiLib/blob/e1c8aece8691222e675233ddfade5485db4ea19b/src/impl/device/motor/motor.cpp#L136

This prevents multiple instances of a `RotationSensor` with conflicting directions from messing each other up. One could even argue having two instances of a `RotationSensor` on the same port with opposite direction is a use-case, on a complex robot with different subsystems controlled by a single axle.

include/okapi/impl/device/distanceSensor.hpp
theol0403: I know you already fixed it, but yeah, I care about that. It gives information at a glance about the purpose of each function and also puts the title in a clear place in the actual documentation section.

dcieslak19973:
> This prevents multiple instances of a `RotationSensor` with conflicting directions from messing each other up. One could even argue having two instances of a `RotationSensor` on the same port with opposite direction is a use-case, on a complex robot with different subsystems controlled by a single axle.

I'm not clear on how this situation could happen, but perhaps I'm not understanding what PROS did under the hood. If I have:
```
RotationSensor rs = RotationSensor(1,false);
RotationSensor ls = RotationSensor(2,true);
```
Won't this currently behave as expected, assuming the sensors are oriented on the right as going forward and the left going reveresed?  As the robot drove straight forward, wouldn't `ls` and `rs` both show positive values?

theol0403:
Yes, that example works fine.

What I am referring to is

```cpp
auto one = RotationSensor(1,false);
auto two = RotationSensor(1,true);
```

Expected: one increases while the other decreases
Actual: in an undefined manner, both of them will show the same value

dcieslak19973:
Makes sense. I was thinking about 2 physical devices which couldn't be plugged into the same port....

Octogonapus:
@theol0403 That is implemented now.

Octogonapus:
@dcieslak19973 Do you have a rotation sensor to test this with? I saw you tested the optical and distance sensors.

dcieslak19973:
I do have 2 physical rotation sensors.  I'll test them physically tonight

theol0403 (dismissed):

dcieslak19973:
Only had 1 rotation sensor; confirmed that 

```
auto one=RotationSensor({3,true});
auto two=RotationSensor({3,false});
```

Behaves as expected - one.get() and two.get() return the same value with opposite signs.

Octogonapus:
> Only had 1 rotation sensor; confirmed that
> 
> ```
> auto one=RotationSensor({3,true});
> auto two=RotationSensor({3,false});
> ```
> 
> Behaves as expected - one.get() and two.get() return the same value with opposite signs.

And the values are in degrees?

dcieslak19973:
Correct, the return values are in degrees for the RotationSensor

Octogonapus (approved):

Issue #433, opened 2020-02-14
AsyncMotionProfileController not respecting gear ratios when based off of a ChassisController


drdeaton:
**Describe the bug**
The AsyncMotionProfileController will turn motors as if the chassis is direct drive green, ignoring any gear ratios you have set in the ChassisController definition.

**Steps To Reproduce**
1. Create a chassis controller with a gear ration (ex: blue cartridge with a 3:5 reduction)

```
auto Chassis = ChassisControllerBuilder()
    .withMotors (
      DriveLeft,
      DriveRight
    )
    .withDimensions(
	  AbstractMotor::gearset::blue,
      {{3.25_in, 26.5_cm}, imev5BlueTPR * (5.0 / 3.0)}
    ).build()
  ;
```

2. Using this ChassisController as a base, create an AsyncMotionProfileController
```
auto ChassisMotionController = AsyncMotionProfileControllerBuilder()
    .withOutput(Chassis)
    .withLimits({1, 2, 10})
    .buildMotionProfileController()
  ;

```

3. Generate and run a path
**Expected behavior**
I expect the chassis to move based on the gear ratios specified in the ChassisControllerBuilder. Instead, it seems to assume green cartridges.

**Versions**
PROS Kernel: 3.2.1
Okapilib: 4.0.3

**Additional Information**
This was tested on a skid-steer chassis with 2 motors driving each side. Each motor turns a 36 tooth gear, which meshes with a 60 tooth gear that is connected to the wheel by quarter inch shaft.
I also don't believe this to be a duplicate of #243, since that issue seemed to be with AsyncMotionProfileController s not based on an existing ChassisController.

drdeaton:
I was able to fix this issue by using the following:

```
  auto Chassis = ChassisControllerBuilder()
    .withMotors (
      DriveLeft,
      DriveRight
    )
    .withDimensions(
	  AbstractMotor::gearset::blue,
      {{3.25_in, 26.5_cm}, imev5BlueTPR * (5.0 / 3.0)}
    ).build()
  ;
  std::cout<<"Built drivebase"<<std::endl;

  auto ChassisMotionController = AsyncMotionProfileControllerBuilder()
    .withOutput(Chassis->getModel(), {{3.25_in, 26.5_cm}, imev5BlueTPR * (5.0 / 3.0)}, {AbstractMotor::gearset::blue, (5.0 / 3.0)})
    .withLimits({.2, .41, 1})
    .buildMotionProfileController()
[...]

PR #461, opened 2020-11-25
Switch from Pathfinder to Squiggles for path generation

baylessj:
### Description of the Change

This PR replaces the dependency on Pathfinder with the Squiggles path generation library.

### Motivation

- Pathfinder is no longer receiving maintenance updates
- Squiggles supports generating paths with negative wheel velocities (needed for tight turns)
- Squiggles has a cleaner API for adding new chassis types like X-drives
- Squiggles is included as a CMake dependency for easier updates and pinning at a specific version

### Possible Drawbacks

The Squiggles library has not yet been tested on a real robot. The library has good test coverage and the paths look correct with its visualization tool but this verification has not been done on a physical robot.

### Verification Process

- Squiggles tests are passing with high code coverage
- Generated a variety of paths and confirmed that they look qualitatively correct with Squiggles path visualization tool
- Okapi tests are passing


.gitignore
```diff
@@ -21,3 +21,5 @@ xml/
 venv/
 *.d/
 build/
+src/squiggles/
+verify-template/
```
Octogonapus: Please put a newline at EOF

firmware/squiggles.mk
```diff
@@ -0,0 +1,3 @@
+DOWNLOAD_SQUIGGLES:=$(shell mkdir -p cmake-build-debug && cd cmake-build-debug && cmake -DCMAKE_BUILD_TYPE=Debug -G "CodeBlocks - Unix Makefiles" .. && mkdir -p ../src/squiggles && find ./squiggles-src/src -type f -name '*.cpp' ! -name 'main.cpp' | xargs cp -t ../src/squiggles && cd ..)
+SQUIGGLES_INC_DIR:=$(ROOT)/cmake-build-debug/squiggles-src/include
+INCLUDE+=-iquote"$(SQUIGGLES_INC_DIR)"
```
Octogonapus: Please put a newline at EOF

src/api/control/async/asyncLinearMotionProfileController.cpp
```diff
@@ -56,84 +56,46 @@ void AsyncLinearMotionProfileController::generatePath(std::initializer_list<QLen
     return;
   }
 
-  std::vector<Waypoint> points;
+  std::vector<squiggles::Pose> points;
   points.reserve(iwaypoints.size());
   for (auto &point : iwaypoints) {
-    points.push_back(Waypoint{point.convert(meter), 0, 0});
+    points.push_back(
+      squiggles::Pose{point.convert(meter), 0, 0});
```
Octogonapus, line 63: Does `squiggles::Pose` accept values in the same frame as `Waypoint`?

src/api/control/async/asyncLinearMotionProfileController.cpp
```diff
@@ -56,84 +56,46 @@ void AsyncLinearMotionProfileController::generatePath(std::initializer_list<QLen
...
   }
 
-  LOG_INFO_S("AsyncLinearMotionProfileController: Preparing trajectory");
+  LOG_INFO_S("AsyncMotionProfileController: Preparing trajectory");
```
Octogonapus:
```suggestion
  LOG_INFO_S("AsyncLinearMotionProfileController: Preparing trajectory");
```

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -57,104 +59,45 @@ void AsyncMotionProfileController::generatePath(std::initializer_list<Pathfinder
     return;
   }
 
-  std::vector<Waypoint> points;
+  std::vector<squiggles::Pose> points;
   points.reserve(iwaypoints.size());
   for (auto &point : iwaypoints) {
     points.push_back(
-      Waypoint{point.x.convert(meter), point.y.convert(meter), point.theta.convert(radian)});
+      squiggles::Pose{point.x.convert(meter), point.y.convert(meter), point.theta.convert(radian)});
```
Octogonapus: Same comment as above about the frame `squiggles::Pose` expects.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -381,60 +322,57 @@ CrossplatformThread *AsyncMotionProfileController::getThread() const {
...
+  internalStorePath(file, ipathId);
 
-  fclose(leftPathFile);
-  fclose(rightPathFile);
+  file.close();
 }
 
 void AsyncMotionProfileController::loadPath(const std::string &idirectory,
                                             const std::string &ipathId) {
+  std::string squigglesPath = makeFilePath(idirectory, ipathId + ".csv");
```
Octogonapus: We should include this new file format in the function's documentation. There should also be a note about loading Pathfinder paths.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -381,60 +322,57 @@ CrossplatformThread *AsyncMotionProfileController::getThread() const {
 
 void AsyncMotionProfileController::storePath(const std::string &idirectory,
                                              const std::string &ipathId) {
-  std::string leftFilePath = makeFilePath(idirectory, ipathId + ".left.csv");
-  std::string rightFilePath = makeFilePath(idirectory, ipathId + ".right.csv");
-  FILE *leftPathFile = fopen(leftFilePath.c_str(), "w");
-  FILE *rightPathFile = fopen(rightFilePath.c_str(), "w");
+  std::string filePath = makeFilePath(idirectory, ipathId + ".csv");
```
Octogonapus, line 327: We should include this new file format in the function's documentation.

src/api/control/async/asyncMotionProfileController.cpp
```diff
@@ -381,60 +322,57 @@ CrossplatformThread *AsyncMotionProfileController::getThread() const {
...
-  // Make sure we can open the file successfully
-  if (leftPathFile == NULL) {
-    LOG_WARN("AsyncMotionProfileController: Couldn't open file " + leftFilePath + " for reading");
-    if (rightPathFile != NULL) {
-      fclose(rightPathFile);
+  
+  std::ifstream leftPathFile, rightPathFile, squigglesPathFile;
+  leftPathFile.open(leftFilePath, std::ifstream::in);
+  rightPathFile.open(rightFilePath, std::ifstream::in);
+  squigglesPathFile.open(squigglesPath, std::ifstream::in);
```
Octogonapus: I would rather that you open the minimum necessary ifstreams. Try squiggles first and close it if it didn't work. Then try pathfinder's next, close them, etc.

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -66,12 +64,349 @@ class AsyncMotionProfileControllerTest : public ::testing::Test {
...
-  FILE *rightPathFile;
-  char *leftFileBuf;
-  char *rightFileBuf;
-  size_t leftFileSize;
-  size_t rightFileSize;
+  std::stringstream squigglesPathFile;
+  std::stringstream leftPathFile;
+  std::stringstream rightPathFile;
+  std::string squigglesFileBuf;
+  const std::string left = "dt,x,y,position,velocity,acceleration,jerk,heading\n"
```
Octogonapus: I would prefer if these strings were moved into separate files.

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -141,11 +476,12 @@ TEST_F(AsyncMotionProfileControllerTest, TwoPathsOverwriteEachOther) {
...
+                                         PathfinderPoint{1_in, 1_in, 180_deg},
                                          PathfinderPoint{3_ft, 0_m, 0_deg},
-                                         PathfinderPoint{3_ft, 1_ft, 0_deg},
-                                         PathfinderPoint{2_ft, 1_ft, 0_deg},
-                                         PathfinderPoint{1_ft, 1_m, 0_deg},
-                                         PathfinderPoint{1_ft, 0_m, 0_deg}},
+                                         PathfinderPoint{3_ft, 1_in, 0_deg},
+                                         PathfinderPoint{3_ft, 1_in, 0_deg},
+                                         PathfinderPoint{1_ft, 1_in, 0_deg},
+                                         PathfinderPoint{1_ft, 0_ft, 0_deg}},
```
Octogonapus: Why does this throw with squiggles?

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -296,8 +632,8 @@ TEST_F(AsyncMotionProfileControllerTest, FollowPathNotMirrored) {
   // Wait a little longer so we get into the path
   rate->delayUntil(200_ms);
 
-  EXPECT_GT(leftMotor->lastVelocity, 0);
-  EXPECT_GT(rightMotor->lastVelocity, 0);
+  EXPECT_NE(leftMotor->lastVelocity, 0);
+  EXPECT_NE(rightMotor->lastVelocity, 0);
```
Octogonapus, line 307: Why was this changed?

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -318,8 +654,8 @@ TEST_F(AsyncMotionProfileControllerTest, FollowPathMirrored) {
...
```
Octogonapus, line 329: Why was this changed?

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -360,16 +696,35 @@ TEST_F(AsyncMotionProfileControllerTest, FilePathRestrict) {
...
-  int genPathLen = controller->getPathData("A").length;
+  int genPathLen = controller->getPathData("A").size();
 
   controller->removePath("A");
-  controller->internalLoadPath(leftPathFile, rightPathFile, "A");
+  controller->internalLoadPath(squigglesPathFile, "A");
   EXPECT_EQ(controller->getPaths().front(), "A");
   EXPECT_EQ(controller->getPaths().size(), 1);
-  EXPECT_EQ(controller->getPathData("A").length, genPathLen);
+  EXPECT_EQ(controller->getPathData("A").size(), genPathLen);
```
Octogonapus: Maybe we can also assert the path data is equal?

test/asyncMotionProfileControllerTests.cpp
```diff
@@ -360,16 +696,35 @@ TEST_F(AsyncMotionProfileControllerTest, FilePathRestrict) {
...
+  leftPathFile.seekg(0);
+  while (!leftPathFile.eof()) {
+    std::getline(leftPathFile, buf);
+    numLines++;
+  }
+  EXPECT_EQ(controller->getPathData("A").size(), numLines - 2);
+
+  controller->setTarget("A");
+  EXPECT_EQ(controller->getTarget(), "A");
+}
```
Octogonapus: Newline at EOF please

Octogonapus (changes requested):

test/asyncMotionProfileControllerTests.cpp
baylessj: That makes sense, they're definitely annoying to keep around in the test source. Think it would be best to give them their own cpp file and create a simple header file to extern them for this test?

test/asyncMotionProfileControllerTests.cpp
baylessj: The Squiggles path generates negative wheel velocities at the start to make a sharp turn. This ensures that we're setting the motors but allows for one of the wheels to be going backwards relative to the path direction (mirrored or not)

test/asyncMotionProfileControllerTests.cpp
baylessj: same as above

src/api/control/async/asyncLinearMotionProfileController.cpp
baylessj: Thanks for catching this, I forgot about that detail. The Squiggles Pose is currently set up to be in cartesian coordinates rather than the ROS-style coordinate frame from Okapi and Pathfinder. I'll translate between these frames

test/asyncMotionProfileControllerTests.cpp
baylessj: The issue comes from the {1_ft, 1_in, 0_deg}->{1_ft, 0_in, 0_deg} segment. It's trying to strafe by an inch basically, and that's too tight of a turn to do properly with the splines. 

Should I add this comment to the test source to clarify it?

src/api/control/async/asyncLinearMotionProfileController.cpp
baylessj: I spoke too soon, the coordinate frames should match between Squiggles and Pathfinder. Yaw is measured counterclockwise from the x-axis like Pathfinder.

test/asyncMotionProfileControllerTests.cpp
Octogonapus: I would prefer if they aren't in a code-type file at all. Just `.txt` files would work fine. You can load from the file during the setup phase of the test suite.

test/asyncMotionProfileControllerTests.cpp
Octogonapus: I see. I have lost track of what all these points do. Let's try to reduce this to the minimum number of points (hopefully just 2). Go ahead and add a comment explaining why the points cause the exception.

baylessj: Couldn't get the strafing path to fail for some reason this time, opted instead to go with a really long path. Comment added specifying the reason for failure.

baylessj:
It's looking like the easiest way to get that test to compile would be to open a PR in your Okapilib Docker image to bump to Ubuntu 20.04 and GCC9 to get the filesystem support. GCC8 has issues with LCOV that prevent it from being a valid option but I've got GCC9 working well for the Squiggles CI. Does that work for you @Octogonapus ?

Octogonapus:
> 
> 
> It's looking like the easiest way to get that test to compile would be to open a PR in your Okapilib Docker image to bump to Ubuntu 20.04 and GCC9 to get the filesystem support. GCC8 has issues with LCOV that prevent it from being a valid option but I've got GCC9 working well for the Squiggles CI. Does that work for you @Octogonapus ?

Not really, I would like to keep the same toolchain. Just use a different solution instead of `<filesystem>`.

baylessj:
Sounds good, switched to using the C API to get the path. It's a bit clunkier but should work.

baylessj:
I just remembered that I'll probably want to pin the squiggles version at a specific number rather than using the latest from the `main` branch. So long as there aren't further changes that need to happen with the Squiggles library for this PR then I can create a 1.0.0 release for it and pin the dependency there

Octogonapus:
Looks good. Just need to run some tests on a real robot now. It would be great if we could see videos of these tests. Not sure if you can do this @baylessj. I can't. Maybe @theol0403 or @dcieslak19973?

dcieslak19973:
I could physically test; do you have an example path in mind? Would you want to demonstrate that paths can have negative wheel velocities? Adding documentation about how to pair this with the OkapiLib velocity PID controller would be nice.

baylessj:
I cannot physically test this, @dcieslak19973 's help would be great. 

I would like to get verification that a path with negative velocities works fine. The exact parameters to make that happen would depend on the track width of your robot, but something where the robot is moving horizontally should work. The points I used to generate the path with negative velocities in the Squiggles README were {0, 0, 10_deg}->{0, 2_m, 20_deg} and a track width of 0.4_m. A similar set of points should get you a nice path; you can also verify the output path with the python visualizer in the Squiggles repo if you'd like.

Besides that, I would recommend testing with any existing paths you have using the current version of Okapilib (if any) or just making some mock auton movements.

baylessj:
The testing should work fine without integrating the AMPC with a velocity PID but I can look into writing a tutorial for that. There's similar content in the [Squiggles docs](https://squiggles.readthedocs.io/en/latest/controller-suggestions.html) but it might be nice to have a similar tutorial in the Okapi docs.

dcieslak19973:
It's certainly possible I've built your branch incorrectly, but I am getting:

```
prosv5 make
Compiled src/main.cpp [WARNINGS]
In file included from ./include/okapi/api.hpp:51:0,
                 from ./include/main.h:42,
                 from ./include/globals.h:1,
                 from ./include/auton.h:1,
                 from src/main.cpp:1:
./include/okapi/api/control/async/asyncLinearMotionProfileController.hpp:18:10: fatal error: squiggles.hpp: No such file or directory
 #include "squiggles.hpp"
          ^~~~~~~~~~~~~~~
compilation terminated.
mv: cannot stat '.d/main.Td': No such file or directory
make: [common.mk:271: bin/main.cpp.o] Error 1 (ignored)
Capturing metadata for PROS Editor...
```

To get to this point, I changed the version in the Makefile to 4.2.0-dev, got the .zip artifact into my test project, ran `pros c fetch okapilib@4.2.0-dev.zip` on it and then `prosv5 c apply okapilib --force-apply --no-download`

baylessj:
> It's certainly possible I've built your branch incorrectly, but I am getting:
> 
> ```
> prosv5 make
> Compiled src/main.cpp [WARNINGS]
> In file included from ./include/okapi/api.hpp:51:0,
>                  from ./include/main.h:42,
>                  from ./include/globals.h:1,
>                  from ./include/auton.h:1,
>                  from src/main.cpp:1:
> ./include/okapi/api/control/async/asyncLinearMotionProfileController.hpp:18:10: fatal error: squiggles.hpp: No such file or directory
>  #include "squiggles.hpp"
>           ^~~~~~~~~~~~~~~
> compilation terminated.
> mv: cannot stat '.d/main.Td': No such file or directory
> make: [common.mk:271: bin/main.cpp.o] Error 1 (ignored)
> Capturing metadata for PROS Editor...
> ```
> 
> To get to this point, I changed the version in the Makefile to 4.2.0-dev, got the .zip artifact into my test project, ran `pros c fetch okapilib@4.2.0-dev.zip` on it and then `prosv5 c apply okapilib --force-apply --no-download`

I'm a bit confused by the error message you've shown, I haven't seen it before. I'll write out the steps I would recommend taking to incorporate my fork into your PROS project; hopefully this will be a good checklist to follow.

- Clone my fork of okapilib (baylessj/okapilib)
- switch to the `feature/squiggles` branch
- in the okapilib root directory, run: 
  - `make`
  - `make template`
  - `prosv5 c f okapilib@*.zip`
- Then, in your project directory, run:
  - `prosv5 c a ../OkapiLib/okapilib@*.zip`
  - `make`

This builds on my machine and hopefully should for yours as well. If it does not, there's a chance that you could be missing a dependency or have a different compiler version. If this fails, please list your gcc version (`arm-none-eabi-gcc --version`) and CMake version (`cmake --version`)

dcieslak19973:
Yeah, that's essentially what I did.

`arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2018-q3-update) 7.3.1 20180622 (release) [ARM/embedded-7-branch revision 261907]`

`cmake version 3.10.2`

In the zip file that `make template` produces, I do not see the squiggles header files. My impression is that they should be in the zip so that pros can unpack them into my robot's code project.

I am able to run the OkapiLib unit tests in the feature branch, which is why I think this is a packaging problem.

baylessj:
I see the issue now, the header files aren't packaged up with the template. They're added to the include path for the Okapilib build but won't be available to a PROS project using the template. I'll fix that.

baylessj:
@dcieslak19973 The latest commit I pushed *should* resolve the issue you found. You'll probably want to bump the okapilib version in your local copy and rebuild the template.

Thanks again for pointing out that error, I wouldn't have caught it otherwise.

dcieslak19973:
Confirmed that my PROS robot project now compiles.  Will physically test later tonight

dcieslak19973:
So the complication here is my test rig has a geared chassis and I'm running into https://github.com/OkapiLib/OkapiLib/issues/433 and the workaround there does not appear to be working for me. So the distance moved by the robot is incorrect, even just for straight line movement.  Not sure if I'll try to resolve that or just deal with the issue

baylessj:
Gotcha, sounds like I'll need to pull in the fix for that to my fork. I can't quite tell from the linked commits if the current state of the #433 branch solves this issue though. Is there additional work that still needs to be done for that issue or can I merge that branch into my fork as-is?

dcieslak19973:
I'll try to do some additional testing today, including reverting to the PathFinder version and see where I get.

Drivetrain is 3.25" wheels geared to speed up by 60 / 36.  My initial test is to move from (0,0,0) to (3ft,0,0). Need to recheck, but I believe the robot only travelled 22 or 25 inches.

dcieslak19973:
Ok, so I'd forgotten what state I'd left things in when I reported above.

With the `squiggles` version installed, the robot doesn't move at all. The terminal log is:

```
5294 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Preparing trajectory
5295 (AsyncMotionProfileController) INFO: Started AsyncMotionProfileController task.
5336 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Completely done generating path A
5336 (User Autonomous (PROS)) DEBUG: AsyncMotionProfileController: Path length: 142
5337 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Set target to: A (ibackwards=0, imirrored=0)
5337 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Waiting to settle
```

I'd then reverted to the `pathFinder` version and got a bit obsessed with the incorrect distance, forgetting that I'd reverted.

When running under the `pathFinder` the terminal log is:
```
11724 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Preparing trajectory
11725 (AsyncMotionProfileController) INFO: Started AsyncMotionProfileController task.
11725 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Generating path
11810 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Modifying for tank drive
11811 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Completely done generating path A
11811 (User Autonomous (PROS)) DEBUG: AsyncMotionProfileController: Path length: 162
11811 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Set target to: A (ibackwards=0, imirrored=0)
11811 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Waiting to settle
11815 (AsyncMotionProfileController) INFO: AsyncMotionProfileController: Running with path: A
11815 (AsyncMotionProfileController) DEBUG: AsyncMotionProfileController: Path length is 162
13435 (AsyncMotionProfileController) INFO: AsyncMotionProfileController: Done moving
13441 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Done waiting to settle
```

baylessj:
Thanks for running that test! The log info is helpful in pointing to the root of the error but I'm not sure what would have caused what you're seeing... The squiggles version seems to get stuck when setting the `isRunning` flag internally, though that hasn't been changed by my fork best I can tell.

How are you starting the move? Are you using discrete generation, target setting, and waiting function calls or using `moveTo()`?

dcieslak19973:
Test harness is (`chassis` is defined elsewhere):

```
  chassis->setState({0_in,0_in,0_deg});
  chassis->setMaxVelocity(75);
  auto profileController = AsyncMotionProfileControllerBuilder()
  .withOutput(chassis)
    .withLimits({1, 2, 10})
    .buildMotionProfileController();

  profileController->generatePath({
      {0_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {3_ft, 0_ft, 0_deg}}, // The next point in the profile, 3 feet forward
      "A" // Profile name
    );
  printf("Setting Target A\n");
  profileController->setTarget("A");
  printf("Set Target A\n");
  profileController->waitUntilSettled();
  printf("Settled\n");
```

I have other routines that use `chassis` such as `chassis->moveDistance(24_in);` which was working.

baylessj:
On a second look at the `AsyncMotionProfileController::loop()`, the one change made in the function may actually be causing this issue. I had run into a segfault in testing when calling `.size()` on the path when the mutex for it hadn't been taken; that might be the issue here too.

@dcieslak19973 can you try commenting out the following line 173/174 of asyncMotionProfileController.cpp?

```
LOG_DEBUG("AsyncMotionProfileController: Path length is " +
                  std::to_string(path->second.size()))
```

It might be good to replace it with similar LOG calls to see where exactly the loop function is failing.

dcieslak19973:
I've commented out the logging lines @baylessj suggested. The thought that this is segfaulting somewhere is very likely. Commenting out that particular log statement did not change the behavior, though I can report additional details.

I still get the same behavior in the logs as before, namely:
```
5294 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Preparing trajectory
5295 (AsyncMotionProfileController) INFO: Started AsyncMotionProfileController task.
5336 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Completely done generating path A
5336 (User Autonomous (PROS)) DEBUG: AsyncMotionProfileController: Path length: 142
5337 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Set target to: A (ibackwards=0, imirrored=0)
5337 (User Autonomous (PROS)) INFO: AsyncMotionProfileController: Waiting to settle
```
Of note, I added additional logging in the `loop` function (inside the `if running` statement) and never see that displayed.

It does appear that the "User Autonomous (PROS)" task dies.  The program appears to be running still, but is no longer responsive to changes in state from the competition switch (these tests are run when I toggle the autonomous to enabled).

The program is definitely crashing inside the `loop` function the first time thru that isRunning is true.

baylessj:
@dcieslak19973 That's helpful debugging info, thanks for passing that along. Unfortunately that report opens up a number of new possibilities for errors that I hadn't considered and I don't have a great answer for you with regards to a solution. 

Tracking this error down is probably a decent bit more hassle than you bargained for, thanks for sticking to this. I think it would be best to figure out why the Autonomous task is dying first; that should point us more clearly to the root of the issue. Can you run the test again but with the following data being sent to the terminal?

- The value of `errno`, a global value that is set with debugging info
- Then when the task dies, can you see if a stack trace is sent over the terminal? This should get you a hex value pointing to the location in the executable that caused the crash and we can use `addr2line` to point that back to the code.

This seems like a possible stack overflow or similar task-specific error to me, but I'm not entirely sure why it would be popping up when reading the `isRunning` value. I expect that the switch to Squiggles has some risks as a result of storing more path data but that data should be going on the heap rather than the autonomous task's stack. 

Now that I mention that, it might actually be worth trying to move the asyncmotionprofilecontroller off the stack. Sorry to ask you to run even more tests, but could you try declaring the AMPC globally and see if that changes the program behavior?

dcieslak19973:
I ended up taking a different approach than your suggestion.  I have isolated the problem within `loop` to the call to `executeSinglePath(path->second, timeUtil.getRate());`.  I've added a `log` line at the start of `executeSinglePath` which does not output to the terminal. Whether that's because the buffer doesn't flush, I'm not sure.  When the `executeSinglePath` line is included, the previous line (`LOG_DEBUG("AsyncMotionProfileController: Path length is " ...)`) does not display to the terminal. When `executeSinglePath` is commented out, this `LOG_DEBUG` statement does show in the terminal.

dcieslak19973:
Ok, a bit more along these lines. Within `executeSinglePath` excluding the code block:
```
    const auto segDT = path[i].time * second;
    const auto leftRPM = convertLinearToRotational(path[i].wheel_velocities[0] * mps).convert(rpm);
    const auto rightRPM =
      convertLinearToRotational(path[i].wheel_velocities[1] * mps).convert(rpm);

    const double rightSpeed = rightRPM / (toUnderlyingType(pair.internalGearset) * pair.ratio) * reversed;
    const double leftSpeed = leftRPM / (toUnderlyingType(pair.internalGearset) * pair.ratio) * reversed;
    if (followMirrored) {
      model->left(rightSpeed);
      model->right(leftSpeed);
    } else {
      model->left(leftSpeed);
      model->right(rightSpeed);
    }
```
Runs.

Adding back in even just the `const auto segDT = path[i].time * second;` line causes it to blow up.

The call to `std::size_t pathSize = path.size(); ` does work, and in my case returns a path size of 142.

baylessj:
Sounds like the `path` object or one of its members is becoming null as we enter the path. I'm not sure why that's happening on your robot but I cannot reproduce it with the unit testing, even when creating paths and running them in the same way.

Can you try and read `errno` and get the stack trace still? I don't know how much that would help with figuring out why the path is NULL but it might be a start.

@Octogonapus does this error seem familiar to you? I'm not sure what would be different between the real brain and our test environment that would cause this error.

dcieslak19973:
> Sounds like the `path` object or one of its members is becoming null as we enter the path. I'm not sure why that's happening on your robot but I cannot reproduce it with the unit testing, even when creating paths and running them in the same way.

I think your theory up the thread about this filling the heap may be accurate. I'll try to rejigger my test harness to make the chassis and motionprofiler global rather than function-scoped and see if that helps.  If it turns out this is the case, I'd suggest we modify the documentation to make clear the safest way to structure the program in terms of global variables.

There was this recent post where it seems like it's possible to alter the memory allocation available to the program running on the V5. I'm not sure how to translate this to the `pros` project: https://www.vexforum.com/t/stack-overflow-in-conways-game-of-life/86888/4

> Can you try and read `errno` and get the stack trace still? I don't know how much that would help with figuring out why the path is NULL but it might be a start.

I can, but I'll admit that I don't know how - any guidance would be appreciated.  I do have a separate thread that runs that logs and can try to add it there, assuming `errno` is global and access is thread-safe.

dcieslak19973:
Ok, significant progress here.  There were 2 bugs in the `executeSinglePath` code which I believe get masked in the unit tests because of the mock implementation of `isSettled()` always returning true (unverified).  Here is the `executeSinglePath` function that I have working:

```
void AsyncMotionProfileController::executeSinglePath(const std::vector<squiggles::ProfilePoint> &path,
                                                     std::unique_ptr<AbstractRate> rate) {
  LOG_INFO_S("AsyncMotionProfileController.executeSinglePath start");
  const int reversed = direction.load(std::memory_order_acquire);
  const bool followMirrored = mirrored.load(std::memory_order_acquire);

  std::scoped_lock pathLock(currentPathMutex);
  // store this locally so we aren't accessing the path when we don't know if it's valid
  std::size_t pathSize = path.size(); 
  currentPathMutex.unlock();
  double prevTime = 0.0;
  for (std::size_t i = 1; i < pathSize && !isDisabled(); ++i) {
    // This mutex is used to combat an edge case of an edge case
    // if a running path is asked to be removed at the moment this loop is executing
    std::scoped_lock lock(currentPathMutex);

    const auto segDT = (path[i].time - prevTime) * second;
    prevTime = path[i].time;

    const auto leftRPM = convertLinearToRotational(path[i].wheel_velocities[0] * mps).convert(rpm);
    const auto rightRPM =
      convertLinearToRotational(path[i].wheel_velocities[1] * mps).convert(rpm);

    const double rightSpeed = rightRPM / (toUnderlyingType(pair.internalGearset) * pair.ratio) * reversed;
    const double leftSpeed = leftRPM / (toUnderlyingType(pair.internalGearset) * pair.ratio) * reversed;

    if (followMirrored) {
      model->left(rightSpeed);
      model->right(leftSpeed);
    } else {
      model->left(leftSpeed);
      model->right(rightSpeed);
    }

    // Unlock before the delay to be nice to other tasks
    currentPathMutex.unlock();

    rate->delayUntil(segDT);
  }
  LOG_INFO_S("AsyncMotionProfileController.executeSinglePath end");
}
```

The first note is that `path[i].time` represents the time-index of the movement - e.g. this is basically an increasing value, not a delta. So I added appropriate logic (feel free to re-write) to get the segment's duration.

The second note is that `rate->delayUntil` does not appear to behave with `0` as an argument. Since `path[0].time` appears to be 0, the first iteration blows up.  I don't know that starting the `  for (std::size_t i = 1; i < pathSize && !isDisabled(); ++i) ` at 1 is "correct", but it causes the robot to at least move.  Alternatively, the `path[0].time` could be a non-zero value.

I also was able to declare my motionProfile variable as a global; I did that first and haven't isolated to see if that had an effect.

dcieslak19973:
An aside in re-reviewing the code above, I think that locking `currentPathMutex` for the whole call might actually lead to fewer potential edge cases than locking/unlocking it every iteration.  It looks like that's been done to handle the edge case where another thread deletes the currentPath. But there's a similar edge case where another thread alters the currentPath, perhaps making it shorter, which would lead to the `for` loop trying to access an element of `path` that no longer exists.

I don't have strong feelings about this, as both edge cases seem like the program is doing something fundamentally dangerous, and would hope few programmers would be manipulating paths in separate threads during path execution.

dcieslak19973:
Using a chassis defined as: `.withDimensions({AbstractMotor::gearset::green,(36.0/60.0)}, {{3.25_in, 13.0_in}, imev5GreenTPR})`, (e.g. an externally geared chassis), the movements go the incorrect distances. I may have time to look into that tonight.

Aside from the distance traveled, the shape of the movement appears reasonable.  The following 2 motion profiles produce the same observed behavior, which is good:

```
  chassis->setState({0_in,0_in,0_deg});
  profileController = AsyncMotionProfileControllerBuilder()
    .withOutput(chassis)
    .withLimits({1, 2, 10})
    .buildMotionProfileController();

  profileController->generatePath({
      {0_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {3_ft, 0_ft, 0_deg}}, // The next point in the profile, 3 feet forward
      "A" // Profile name
    );
  profileController->setTarget("A");
  profileController->waitUntilSettled();
```

And:
```
  chassis->setState({1_ft,0_in,0_deg});
  profileController->generatePath({
      {1_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {4_ft, 0_ft, 0_deg}}, // The next point in the profile, 3 feet forward
      "B" // Profile name
    );
  profileController->setTarget("B");
  profileController->waitUntilSettled();
```

The following produces an S-shaped movement consistent with the documentation:
```
  profileController = AsyncMotionProfileControllerBuilder()
    .withOutput(chassis)
    .withLimits({1, 2, 10})
    .buildMotionProfileController();

  profileController->generatePath({
      {0_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {4_ft, 4_ft, 0_deg}},
      "C" // Profile name
    );
  profileController->setTarget("C");
  profileController->waitUntilSettled();
```

The following produces a runtime exception. I believe the path generates but it fails in waitUntilSettled:
```
  chassis->setState({0_in,0_in,0_deg});
  profileController = AsyncMotionProfileControllerBuilder()
    .withOutput(chassis)
    .withLimits({1, 2, 10})
    .buildMotionProfileController();

  profileController->generatePath({
      {0_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {0_ft, 4_ft, 0_deg}}, 
      "D" // Profile name
    );
  profileController->setTarget("D");
  profileController->waitUntilSettled();
```

baylessj:
Thank you for the debugging, @dcieslak19973, you've been a huge help! I will integrate your fixed `executeSinglePath` tomorrow into my fork.

dcieslak19973:
I haven't fully debugged things yet. One thing to be aware of is that because `executeSinglePath` winds up calling `model->left(leftSpeed)` this is influenced by `model->getMaxVelocity()` given that `model->left()` winds up calling something like:

```
void SkidSteerModel::left(const double ispeed) {
  leftSideMotor->moveVelocity(static_cast<int16_t>(std::clamp(ispeed, -1.0, 1.0) * maxVelocity));
}
```

It may be the case that we would want `executeSinglePath` to set the model's maxVelocity at the beginning and then restore it's value after the movement has completed.

baylessj:
Resolved the issue with the delay that you mentioned. I'd like to make sure I understand your other points correctly, though:

- I don't think we can lock the mutex for the duration of the path. The mutex gets released to allow for the possibility of safely deleting paths when in the middle of running them. I used this approach to exit a path on an exit condition in my Turning Point robot.
- I'm a bit confused by why path "D" would be failing in your example paths above... Can you try to identify where in `waitUntilSettled` you're getting an error? It might also be helpful to print some confirmation at the end of the path generation to verify that it is actually generating right.
- Are you modifying the maximum velocity for your chassis in the middle of the path? I'm not sure why it would be an issue that the output of the AMPC uses the chassis' maximum velocity
- This was discussed a little while back, but is there anything that needs to be done on this PR to fix your issue with the drivetrain gear ratio?

dcieslak19973:
> * I don't think we can lock the mutex for the duration of the path. The mutex gets released to allow for the possibility of safely deleting paths when in the middle of running them. I used this approach to exit a path on an exit condition in my Turning Point robot.

Fair.

> * I'm a bit confused by why path "D" would be failing in your example paths above... Can you try to identify where in `waitUntilSettled` you're getting an error? It might also be helpful to print some confirmation at the end of the path generation to verify that it is actually generating right.

Will do

> * Are you modifying the maximum velocity for your chassis in the middle of the path? I'm not sure why it would be an issue that the output of the AMPC uses the chassis' maximum velocity

Here's the code in my test-suite:
```
  chassis->setState({0_in,0_in,0_deg});
  chassis->setMaxVelocity(75);

  printf("chassis->gearRatio = %.1f\n",chassis->getGearsetRatioPair().ratio);
  profileController = AsyncMotionProfileControllerBuilder()
    .withOutput(chassis)
    .withLimits({1, 2, 10})
    .buildMotionProfileController();

  profileController->generatePath({
      {0_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {3_ft, 0_ft, 0_deg}}, // The next point in the profile, 3 feet forward
      "A" // Profile name
    );
  printf("Setting Target A\n");
  profileController->setTarget("A");
  printf("Set Target A\n");
  profileController->waitUntilSettled();
  printf("Settled\n");
```

Which leads to the robot travelling the incorrect distance.  This change to `executeSinglePath` appears to correct for it:

```
void AsyncMotionProfileController::executeSinglePath(const std::vector<squiggles::ProfilePoint> &path,
                                                     std::unique_ptr<AbstractRate> rate) {
  const int reversed = direction.load(std::memory_order_acquire);
  const bool followMirrored = mirrored.load(std::memory_order_acquire);
  double priorMaxVel = model->getMaxVelocity();
  model->setMaxVelocity(toUnderlyingType(pair.internalGearset));

  currentPathMutex.lock();
  // store this locally so we aren't accessing the path when we don't know if it's valid
  std::size_t pathSize = path.size(); 
  currentPathMutex.unlock();
  for (std::size_t i = 0; i < pathSize && !isDisabled(); ++i) {
    // This mutex is used to combat an edge case of an edge case
    // if a running path is asked to be removed at the moment this loop is executing
    std::scoped_lock lock(currentPathMutex);

    const auto segDT = DT * second;
    const auto leftRPM = convertLinearToRotational(path[i].wheel_velocities[0] * mps).convert(rpm);
    const auto rightRPM =
      convertLinearToRotational(path[i].wheel_velocities[1] * mps).convert(rpm);

    const double rightSpeed = rightRPM / toUnderlyingType(pair.internalGearset) * reversed;
    const double leftSpeed = leftRPM / toUnderlyingType(pair.internalGearset) * reversed;

    LOG_DEBUG("AsyncMotionProfileController.executePath path[" + std::to_string(i) + "]=" +
      path[i].to_string());

    LOG_DEBUG("AsyncMotionProfileController.executePath path[" + std::to_string(i) + "].lRPM=" + std::to_string(leftRPM) + " rRPM=" + std::to_string(rightRPM));
    LOG_DEBUG("AsyncMotionProfileController.executePath path[" + std::to_string(i) + "].lSpeed=" + std::to_string(leftSpeed) + " rSpeed=" + std::to_string(rightSpeed));

    if (followMirrored) {
      model->left(rightSpeed);
      model->right(leftSpeed);
    } else {
      model->left(leftSpeed);
      model->right(rightSpeed);
    }

    // Unlock before the delay to be nice to other tasks
    currentPathMutex.unlock();

    rate->delayUntil(segDT);
  }
  model->setMaxVelocity(priorMaxVel);
}
```

Though it is possible the last line does not get called in all cases.

> * This was discussed a little while back, but is there anything that needs to be done on this PR to fix your issue with the drivetrain gear ratio?

I'm reasonably sure that the distance issue is solved with the `setMaxVelocity` change above.  I think I was incorrect in attributing the improper distance travelled with the external gearing.  That said, this PR https://github.com/OkapiLib/OkapiLib/pull/464 does address some issues I found with external gearing.

dcieslak19973:
With the change posted above to `executeSinglePath` the robot moves generally as expected. I've not been able to reproduce the error state that I had previously observed.

Depending on your and @Octogonapus thoughts on the `chassis->setMaxVelocity` issue, I would be in favor of approving this PR.

dcieslak19973:
In playing around a little more, I think the Squiggles co-ordinates and the OkapiLib Odom co-ordinates are possibly mirrored.

baylessj:
You're right, +y is left of the robot in squiggles rather than +y being right for okapi before. I'll add a negation to the Y axis to correct for this.

I think the max velocity issue should be handled in a separate PR if that's alright.

dcieslak19973:
Will test new change; probably won't be able to until Monday.

I'm good with the max velocity in a separate PR; maybe create an issue for it so we remember

dcieslak19973:
I tested your inversion-of-y patch and it's still not quite right. Changing that line instead to:

```
squiggles::Pose{point.y.convert(meter),
                      point.x.convert(meter),
                      (90_deg - point.theta).convert(radian)});
```

Produces better results. I'm not confident that the transform on theta is correct, though it appears to work for my physical test.  I also tried to constrain the theta conversion to -180 to 180 (or 0-360) but couldn't get the `OdomMath.constrainAngle180` function to work inside the constructor

dcieslak19973:
@baylessj ^

baylessj:
These sorts of state transformations are always confusing to me, I'll try and clarify the details I do know first though.

[Okapi's coords](https://okapilib.github.io/OkapiLib/namespaceokapi.html#af37fbd761bd859a00ff4dd4a87dd8c07ad5ed7666e5cebf60d3af20a5a46edf3b) are +x forward, +y right, 0 theta on the x axis, and +theta going from x to y (clockwise).

Squiggles' coords are traditional cartesian, +x is forward, +y is left, 0 theta is on the x axis, and +theta is going from x to y (counter-clockwise).

The Squiggles' definition of Cartesian differs from the version of Cartesian used by the Okapi Odometry, where +x is right, +y is forward, 0 theta is on the Y axis, and +theta goes from x to y (counter-clockwise).

I was confused in my prior comment and thought that Okapilib used the [ROS coords](https://www.ros.org/reps/rep-0103.html) where +x is forward, +y is left, 0 theta is on the x axis, and +theta is going from x to y (counter-clockwise). This matches the Squiggles coordinate frame.

I looked at the frame transformation in the `twoEncoderOdometry.cpp` file as a reference and I'm not sure that the transformation there is correct... The transformation from Okapi_Standard->Okapi_Cartesian just flips x and y, keeping theta the same. This does not line up with my earlier statement that the two frames use opposing +theta values.

I believe, based on the above information, that right frame transformation from Okapi_Standard to Squiggles is the following:

```
squiggles::Pose{point.x.convert(meter),
                      (-1 * point.y).convert(meter),
                      (-1 * point.theta).convert(radian)});
```

@dcieslak19973 Does that transformation make sense to you? If so, could you try it on your robot to confirm?

dcieslak19973:
@baylessj - I agree, the co-ordinate transformations are not pleasant.

I'll try what you propose above later tonight, namely:
```
squiggles::Pose{point.x.convert(meter),
                      (-1 * point.y).convert(meter),
                      (-1 * point.theta).convert(radian)});
```

I may also change my test program to to switch to `CARTESIAN` `.withOdometry` configuration in the chassis builder and see where that goes.

It's possible that the `twoEncoderOdometry.cpp` transformation is incorrect. Perhaps @theol0403 or @Octogonapus could weigh in on it.

dcieslak19973:
@baylessj 

I switched my test program to use OkapiLib CARTESIAN and changed my path-points passed into the motionprofiler to correspond with those.  The different co-ordinate conventions are making my head spin.  In `AsyncMotionProfileController::generatePath`, using this snippet:

```
  std::vector<squiggles::Pose> points;
  points.reserve(iwaypoints.size());
  for (auto &point : iwaypoints) {
    points.push_back(
      squiggles::Pose{point.x.convert(meter), point.y.convert(meter), (90_deg - point.theta).convert(radian)});
  }
```

Produces the desired movement.  I did try `squiggles::Pose{point.x.convert(meter), -1.0 * point.y.convert(meter), -1.0 * point.theta.convert(radian)});` as well as a couple other permutations which would generally produce counter-clockwise movement when clockwise movement was expected.

During the movements, I had a separate thread output the OkapiLib (Cartesian) co-ordinates/heading which matched expectation.  My hope is that this work now will allow for an easier time should one decide to try to implement a pure-pursuit path follower down the road.

The change listed above does make the unit tests on the AsyncMotionProfileControllerTest.FollowPath* fail.  I did not physically test reversed and mirrored paths.

baylessj:
@dcieslak19973 The transform you listed above differs slightly from the one that you posted before, you had x and y flipped originally but not in the second example. I think your first transformation should work. For reference I'm going to compare a vector in each coordinate system:

   ^       60    /         ^                        ^
   |              /            |                         |
1 |            x             |  X    (Okapi)     | Y    (Squiggles)
   |                           |                         |
   ------------>           --------->             ---------->
                2                      Y                         X   

Yaw is measured from X to Y in both frames.

Using the above point as an example, we should have {1, 2, 60} in the Okapilib `FRAME_TRANSFORMATION` frame. This should become {2, 1, 30} in the Squiggles form.  

This looks to me like your first transformation would work then. Can you try that for an end point in each quadrant and verify that it works? (Like {2, 1, 30}, {-2, 1, -30}, {2, -1, 120}, {-2, -1, -120}, for example)

dcieslak19973:
@baylessj - that's correct. The first transform which flips `x` and `y` works assuming coordinates are sent in using OkapiLib.FRAME_TRANSFORMATION (which appears to be the the default for the chassis builder), while the second transform works assuming OkapiLib.CARTESIAN.

I think the "correct" solution would be for the motion profiler to handle both cases. However, I think that could be done as a follow on PR, as I think it would potentially change the method signature (to accept the new parameter defaulted to FRAME_TRANSFORMATION), or to infer it based on the chassis controller.  But the motion profilers seem to take `chassisModel` and not `chassisController` in their constructors (which I'm not quite sure what the rationale there is).

TL;DR - in the short term it's probably sufficient to implement:

```
squiggles::Pose{point.y.convert(meter),
                      point.x.convert(meter),
                      (90_deg - point.theta).convert(radian)});
```
With the suggestion that the comments make clear that the waypoints passed in are expected to be OkapiLib.FRAME_TRANSFORMATION format.

baylessj:
That makes sense to me, thanks for the clarification. I'll implement the transformation that you have listed above and specify the frame transformation in the docs. This will match with the current behavior like I'm intending; the goal is to not create any breaking changes here.

dcieslak19973:
Nice. Thanks @baylessj 

@theol0403 / @Octogonapus - I'm a +1

dcieslak19973:
@theol0403 / @Octogonapus - bump

Octogonapus:
So the last details have been worked out and tested on a real robot then?

Thank you for your work on this @dcieslak19973 and @baylessj

dcieslak19973:
Yes; physically tested on a real robot

Octogonapus:
@dcieslak19973 Update: @baylessj and I talked offline and we are going to wait on this PR until @baylessj finishes some changes in Squiggles.

baylessj:
The issues blocking the PR mentioned above have been resolved. I'd like the code to be retested but then it should be good to go.

@dcieslak19973 Would you be able to repeat the test cases you had run before but with the latest changes to this PR?

dcieslak19973:
@baylessj - I should be able to take a look at it sometime in the next week

dcieslak19973:
I've confirmed that it works as it did before.  I did throw one potential edge case at it which looks to generate a runtime exception (I think it was an impossible path exception):

```
  profileController->generatePath({
      {0_ft, 0_ft, 0_deg},  // Profile starting position, this will normally be (0, 0, 0)
      {0_ft, 0_ft, 90_deg}}, // Point turn - this may be unfair
      "A" // Profile name
    );
  printf("Setting Target A\n");
  profileController->setTarget("A");
  printf("Set Target A\n");
  profileController->waitUntilSettled();
  printf("Settled\n");
```

baylessj:
Thank you for testing this! I'm not surprised that the point turn fails, although now that you bring it up it might be a nice option to support point turns in the future. I'll open an issue for that on the squiggles repo and look into supporting it in the future.

PR #464, opened 2020-11-29
use odom gearset/drivescales when available

### LINKED ISSUE #462: ChassisControllerBuilder::buildCCPID() should use odomScales when set
**Describe the bug**
When differentOdomScales is true, buildCCPID() should use odomScales instead of driveScales.

**Steps To Reproduce**
Create a chassis with an externally geared drive ratio and "vanilla" tracking wheels.  Observe that when moving, the movement will be incorrect.

**Expected behavior**
Movements should be correct regardless of whether the drivetrain and odometry have different gearing ratios

**Versions**
4.1.0

**Additional Information**
N/A

dcieslak19973:
### Description of the Change

When using geared drivetrains and odom, the odom drivescales are not used, leading to inaccurate movements

### Motivation

Correct behavior described above

### Possible Drawbacks

N/A

### Verification Process

Physical test

### Applicable Issues

Fixes #462

### DIFF
--- a/src/api/chassis/controller/chassisControllerPid.cpp
+++ b/src/api/chassis/controller/chassisControllerPid.cpp
@@ -118,6 +118,7 @@ void ChassisControllerPID::trampoline(void *context) {
 
 void ChassisControllerPID::moveDistanceAsync(const QLength itarget) {
   LOG_INFO("ChassisControllerPID: moving " + std::to_string(itarget.convert(meter)) + " meters");
+  LOG_DEBUG("ChassisControllerPID: straight " + std::to_string(scales.straight) + " ratio " + std::to_string(gearsetRatioPair.ratio));
 
   distancePid->reset();
   anglePid->reset();
@@ -126,7 +127,7 @@ void ChassisControllerPID::moveDistanceAsync(const QLength itarget) {
   turnPid->flipDisable(true);
   mode = distance;
 
-  const double newTarget = itarget.convert(meter) * scales.straight * gearsetRatioPair.ratio;
+  const double newTarget = itarget.convert(meter) * scales.straight;
 
   LOG_INFO("ChassisControllerPID: moving " + std::to_string(newTarget) + " motor ticks");
 
@@ -155,6 +156,7 @@ void ChassisControllerPID::moveRaw(const double itarget) {
 void ChassisControllerPID::turnAngleAsync(const QAngle idegTarget) {
   LOG_INFO("ChassisControllerPID: turning " + std::to_string(idegTarget.convert(degree)) +
            " degrees");
+  LOG_DEBUG("ChassisControllerPID: straight " + std::to_string(scales.turn) + " ratio " + std::to_string(gearsetRatioPair.ratio));
 
   turnPid->reset();
   turnPid->flipDisable(false);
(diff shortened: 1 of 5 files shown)


theol0403:
Fyi: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

src/impl/chassis/controller/chassisControllerBuilder.cpp
```diff
@@ -413,6 +413,12 @@ ChassisControllerBuilder::buildDOCC(std::shared_ptr<ChassisController> chassisCo
 }
 
 std::shared_ptr<ChassisControllerPID> ChassisControllerBuilder::buildCCPID() {
+  auto ccGearset = gearset;
+  auto ccDriveScales = driveScales;
+  if(differentOdomScales) {
+    ccDriveScales = odomScales;
```
Octogonapus: Can't we just use `odomScales` in either case?

src/impl/chassis/controller/chassisControllerBuilder.cpp
```diff
@@ -413,6 +413,12 @@ ChassisControllerBuilder::buildDOCC(std::shared_ptr<ChassisController> chassisCo
...
+    ccGearset = AbstractMotor::gearset::green;
```
Octogonapus: Why hard-code green?

Octogonapus (changes requested):

src/impl/chassis/controller/chassisControllerBuilder.cpp
dcieslak19973: Will test whether that works.

src/impl/chassis/controller/chassisControllerBuilder.cpp
dcieslak19973: Yeah, that may not be necessary. will test

dcieslak19973: @Octogonapus - During physical testing I ran into issues with an externally geared chassis which I've also resolved in this PR. Feedback?

src/api/chassis/controller/chassisControllerPid.cpp
```diff
@@ -126,7 +127,7 @@ void ChassisControllerPID::moveDistanceAsync(const QLength itarget) {
   turnPid->flipDisable(true);
   mode = distance;
 
-  const double newTarget = itarget.convert(meter) * scales.straight * gearsetRatioPair.ratio;
+  const double newTarget = itarget.convert(meter) * scales.straight;
```
Octogonapus: Is this change (and the one below) backwards compatible for people that don't call `withDimensions` with a `GearsetRatioPair`?

dcieslak19973: Yes, if one does `.withDimensions(AbstractMotor::gearset::green, {{4.0_in,10_in}, imev5GreenTPR})` it will create with a gearsetRatioPair with a 1.0 gear ratio defaulted.

Octogonapus: I am not sure where the external ratio is accounted for, then. It seems like it is always discarded.

dcieslak19973: @Octogonapus - you are correct. Latest commit should fix this. The chassisControllerPID now works physically with both external tracking wheels and configured to use the drive train as the odom source.

I've phyically tested this with both configurations

Octogonapus (approved):
Thanks for working on this rats nest :+1:

Issue #303, opened 2019-01-07
Iterative Chassis PID - Give the chassis some error and it will do 1 step of PID


theol0403:
Currently, there are two pairs of movement functions.
`moveDistance`/`moveDistanceAsync` and `turnAngle`/`turnAngleAsync`.
These each set the target of the chassis relative to current position then execute the motion.
However, I want to be able to define my own error calculation for the chassis to do PID on.
What I mean is to be able to use the chassis in a way similar to an `IterativePIDController` where you can specify the target and then step. However, the implementation might be different for a chassis.

This will be useful for

- Active Brake
- Alignment (Such as Vision Sensor Alignment)
- Angle/Point Seeking

Lets say for example we have Odometry. We are able to read the absolute chassis angle at any time.
Using the above functions, if we wanted to turn to face 0 degrees, we would have to do something along the lines of
`robotChassis.turnAngle(0_deg - chassisOdom.theta);`
If chassis theta is -90 degrees, this will then set the target of the chassis to +90 degrees.
The robot will then turn to face 0 degrees.
However, if the robot slips, or something goes wrong with the turn, the target will become messed up and the turn will be inaccurate. To attempt to hone in to the 0 degrees, you would have to run the function again and hope that it will reach the target.

However, all this time the chassisTheta was kept accurate.
This is the request - to be able to specify the value to calculate the error from.

This is then what you could do to hone in to 0 degrees.
```cpp
do
{
    robotChassis.turnAngleIterative(0_deg - chassisOdom.theta); // Does 1 step of PID from specified error
    pros::delay(20);
}
while (!robotChassis.isSettled);
```
This way it does PID on the calculation we used to determine the wanted angle instead of calculating once and then doing PID on that static target.
Another way to do this would be
```cpp
while(true)
{
    if(myButton)
    {
        //Does 1 step of PID given error
        // could be any arbitrary calculation giving turning error
        myChassis->turnAngleIterative(targetAngle - currentAngle); 
        pros::delay(20);
    }
}
```

Then it will only do PID when the button is pressed.
This might be useful for vision alignment or anything where you want the chassis to use PID to react to external values. 
You would also be able to specify a dynamic variable as the target and move the target around.
So as a recap, instead of calculating and executing the whole movement as one action, it instead does one step of PID towards the target and then re-calculates the target.
I think the best way of doing this is by manually calculating the error, but there may be other ways such as setting the target with an input.

dcieslak19973:
[...]

PR #465, opened 2020-12-24
First pass at the API

dcieslak19973:
Have not tested

### Description of the Change

Add `moveDistanceIterative` and `turnAngleIterative` to chassis classes.

See:

https://github.com/OkapiLib/OkapiLib/issues/303

### Motivation

This will be useful for

* Active Brake
* Alignment (Such as Vision Sensor Alignment)
* Angle/Point Seeking

For example:
```
do
{
    robotChassis.turnAngleIterative(0_deg - chassisOdom.theta); // Does 1 step of PID from specified error
    pros::delay(20);
}
while (!robotChassis.isSettled());
```

### Possible Drawbacks

Unclear what the correct behavior for the Integrated chassis type is.  Currently, this implementation passes it down to the corresponding `Async` functions. Could easily throw an exception instead.

### Verification Process

Not yet verified. First commit is to make sure there is alignment on the API. Rough cut at the implementation details.

### Applicable Issues

https://github.com/OkapiLib/OkapiLib/issues/303


dcieslak19973:
@theol0403 - thoughts?

dcieslak19973:
@theol0403 / @Octogonapus - bump

Octogonapus:
In thinking more about the motivation, I think that all that is really required from the implementation is a method similar to `moveDistanceAsync` that does not reset the PID controllers. Using a method like that, you could walk the target of the PID controllers as needed to track a reference. Selective pausing the PID loop like Theo mentions in #303 could be implemented using the existing `stop` method. The body of `moveDistanceAsync` (after the calls to `reset` and before setting `newMovement`) could be refactored out into `moveDistanceIterative`. The same goes for `turnAngleAsync`.

dcieslak19973:
This is still physically untested.

@Octogonapus  - thanks for the feedback.  I'm not sure that I could exactly refactor as you suggest; I don't think that `stop` would actually pause the PID loop as you may expect - it winds up calling `chassis->stop()` when it gets into `stopAfterSettled()`.

This refactor at least eliminates some duplicated code and has the PID loop actually call "IterativeStep" (protected) functions that the Iterative functions also call, which seems like it adheres to the spirit of this feature.

I thought about providing public `resetForDistanceIterative()` and `resetForTurnIterative()` functions that would reset the PIDs, and the doneLooping/newMovement flags, but that seems like a user might not do right.

dcieslak19973:
This is in a good state for either approval or feedback for changes

dcieslak19973:
@Octogonapus - Any feedback?  If not, good to merge?

Octogonapus:
Closed due to me deleting develop. I would rebase on master but it looks like you didn't enable edits.

I have more time to review this now.

PR #471, opened 2021-03-30
Feature add chassis motor rate limits

Sleuth56:
### Description of the Change
In the chassisPID controller we use open loop control to drive the motors calculating the motor power values on each cycle. So by adding a limiter to those values this allows for the input to very quite heavily but the motors only change a little at a time. For example when the PID tells the motors to go to max power from a standstill.

### Motivation
My team's robot is quite light as many team's are this season. So the front would buck when driving larger distances. This fixes that by limiting the rate of change of the drive train motors.

### Possible Drawbacks
When switching from PID to integrated will cause an error if withmaxRates are defined as it's not possible to support the integrated PID's do to them being controlled in a closed loop fashion.

### Verification Process
I've been running this on my team's competition robot for 2-3 weeks now. Including one competition. But I haven't had time to test all possible configurations. Example with h-drive or mecanum drive trains.

Thanks to @johnholbrook who helped me get through some C++ elements I didn't understand.


Octogonapus:
I am not sure I want this added. If more people in the community come forward in support of this feature, I will be on board with adding it.

Sleuth56:
I know that there is some level of interest in this. What's the reason to not add this feature? Was my implementation correct or is there a better way?

Octogonapus:
Your implementation needs work, yes, but that is not the point. I don't have time to maintain this project as it is; more features are more for me to maintain.

Octogonapus:
closing because of the prior reasons and because this is quite stale now

PR #478, opened 2022-05-22
Add the 'tile' unit of QLength

karmanyaahm:
I don't know if you guys want this, but it's a pretty small change if you do:

### Description of the Change

Add a unit called 'tile' = 24 inches. Just like length values to chassis PID and other places can be 24_in, it can be done with 1_tile.

### Motivation

The VRC field is 6 by 6 tiles. This could be convenient for teams coding their auton movements relative to a certain number of field tiles rather than just inches.

See this VexForum post: https://www.vexforum.com/t/tipping-point-season-code-request/102621/8?u=karmanyaahm .

### Possible Drawbacks

I don't think there's any since it's not removing or changing anything

### Verification Process

I've been using an equivalent of this in my competition code all year, and this is a pretty small change.

Another cool thing I do in my code, is printing odometry values (OdomState::str) as _tile instead of meter. 6 tiles is more intuitive than seeing 3.6576 m (the field dimensions). However, I'm unsure of how to implement that without any regressions, if you guys think it's a good idea, I can work on it with some advice.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,4 +7,4 @@
 	url = https://github.com/eranpeer/FakeIt.git
 [submodule "m.css"]
 	path = m.css
-	url = git://github.com/mosra/m.css
+	url = https://github.com/mosra/m.css.git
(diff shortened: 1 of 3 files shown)


Octogonapus (approved):
LGTM

PR #484, opened 2022-05-27
Allow IMU to reset to custom angles

### LINKED ISSUE #470: Allow IMU to reset to angles other than 0
When resetting the IMU, users are only able to reset the angle to zero degrees. This is problematic when the user is aligning with some object which is not at 0 degrees; for example, a side wall, where the angle should be 90 degrees

Allow the user to supply an optional new angle in the ``reset()`` function and set the offset accordingly. The new function would look like this:
```cpp
std::int32_t IMU::reset(const double inewAngle=0) {
  currentReading = readAngle();
  if (currentReading == PROS_ERR) {
    return PROS_ERR;
  } else {
    offset = currentReading - inewAngle;
    return 1;
  }
}
```
If the imu currently reads 92 and the user wishes to set the angle to be 90, the offset would then be 2 degrees. I'm pretty sure the signs are correct here, but obviously it should be tested to make sure.

A similar change could be made to the ``calibrate()`` function

Users can manually offset the returned value or avoid resetting to different angles.

**Versions**
PROS: 3.3.1
Okapi: 4.2.0

jazonshou:
### Description of the Change

This feature allows users to reset the IMU angle to angles other than 0. 

### Motivation

It did not exist before. 

### Possible Drawbacks

None, as of right now

### Verification Process

I checked the math and it should work. 

### Applicable Issues

Closes #470

### DIFF
--- a/include/okapi/impl/device/rotarysensor/IMU.hpp
+++ b/include/okapi/impl/device/rotarysensor/IMU.hpp
@@ -63,9 +63,9 @@ class IMU : public ContinuousRotarySensor {
   std::int32_t reset() override;
   
   /**
-   * @brief Resets rotation value to desired value
-   *        For example, ``reset(0)`` will reset the sensor to zero.
-   *        But ``reset(90)`` will reset the sensor to 90 degrees.
+   * Resets rotation value to desired value
+   * For example, ``reset(0)`` will reset the sensor to zero.
+   * But ``reset(90)`` will reset the sensor to 90 degrees.
    * 
    * @param inewAngle desired reset value
    * @return ``1`` or ``PROS_ERR``.
--- a/include/okapi/impl/device/rotarysensor/IMU.hpp
+++ b/include/okapi/impl/device/rotarysensor/IMU.hpp
@@ -61,6 +61,16 @@ class IMU : public ContinuousRotarySensor {
    * @return ``1`` or ``PROS_ERR``.
    */
   std::int32_t reset() override;
+  
+  /**
+   * @brief Resets rotation value to desired value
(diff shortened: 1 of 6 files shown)


include/okapi/impl/device/rotarysensor/IMU.hpp
```diff
@@ -61,6 +61,16 @@ class IMU : public ContinuousRotarySensor {
    * @return ``1`` or ``PROS_ERR``.
    */
   std::int32_t reset() override;
+  
+  /**
+   * @brief Resets rotation value to desired value
```
Octogonapus: Don't use `@brief`

src/impl/device/rotarysensor/IMU.cpp
```diff
@@ -56,6 +56,16 @@ std::int32_t IMU::reset() {
   }
 }
 
+std::int32_t IMU::reset(double inewAngle) {
+  double currentAngle = readAngle();
+  if(currentAngle = PROS_ERR) {
```
Octogonapus:
```suggestion
  if (currentAngle == PROS_ERR) {
```

src/impl/device/rotarysensor/IMU.cpp
```diff
@@ -56,6 +56,16 @@ std::int32_t IMU::reset() {
   }
 }
 
+std::int32_t IMU::reset(double inewAngle) {
+  double currentAngle = readAngle();
```
Octogonapus:
```suggestion
  const double currentAngle = readAngle();
```

Octogonapus (changes requested):

Octogonapus (approved):

Issue #468, opened 2021-03-01
Make TurnToAngle take the shortest angle


dcieslak19973:
**Describe the bug**
Current call uses:

`  const auto angle = iangle - odom->getState(StateMode::FRAME_TRANSFORMATION).theta;`

Which can lead to 270 degree turns instead of 90 degree turns.

Changing to:

`  const auto angle = OdomMath::constrainAngle180(iangle - odom->getState(StateMode::FRAME_TRANSFORMATION).theta);`

Should ensure the robot turns the shortest angle 

**Steps To Reproduce**

Initialize robot to a zero degree heading and ask it to TurnToAngle(-90_deg);

**Expected behavior**
Instead of turning clockwise 270 degrees, robot should turn 90 degrees counter-clockwise

**Versions**
Current

**Additional Information**
Good first issue for someone

jazonshou:
This issue can be closed.

PR #485, opened 2022-06-01
Turns shortest angle in OdomChassisController

jazonshou:
### Description of the Change

OdomChassisController will now turn the shortest turn when ``turnToAngle`` is called. 

### Motivation

Currently, turnToAngle will only turn clockwise. For example, if I wanted to turn to 270_deg, it would turn 270 degrees clockwise instead of turning 90 degrees counterclockwise. 

### Possible Drawbacks

None that I can think of right now 

### Verification Process

The ``turnToPoint`` function uses a similar method of constraining the desired turn angle to [-180, 180], and it works, so this change for ``turnToAngle`` should work too. 

### Applicable Issues

#468

### DIFF
--- a/include/okapi/api/chassis/controller/chassisController.hpp
+++ b/include/okapi/api/chassis/controller/chassisController.hpp
@@ -57,6 +57,7 @@ class ChassisController {
 
   /**
    * Turns the robot clockwise in place (using closed-loop control).
+   * (turns the shortest angle for OdomChassisControllers)
    *
    * @param idegTarget angle to turn for
    */
@@ -71,6 +72,7 @@ class ChassisController {
 
   /**
    * Sets the target angle for the robot to turn clockwise in place (using closed-loop control).
+   * (turns the shortest angle for OdomChassisControllers)
    *
    * @param idegTarget angle to turn for
    */
--- a/include/okapi/api/chassis/controller/chassisController.hpp
+++ b/include/okapi/api/chassis/controller/chassisController.hpp
@@ -57,7 +57,6 @@ class ChassisController {
 
   /**
    * Turns the robot clockwise in place (using closed-loop control).
-   * (turns the shortest angle for OdomChassisControllers)
    *
    * @param idegTarget angle to turn for
    */
@@ -72,7 +71,6 @@ class ChassisController {
 
   /**
    * Sets the target angle for the robot to turn clockwise in place (using closed-loop control).
-   * (turns the shortest angle for OdomChassisControllers)
    *
    * @param idegTarget angle to turn for
    */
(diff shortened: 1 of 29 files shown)


Octogonapus:
Please allow edits from maintainers.

jazonshou:
> Please allow edits from maintainers.

They should be allowed....

Octogonapus:
I get this error when pushing to your branch
```
To github.com:Ryan4253/OkapiLib.git
 ! [remote rejected]   HEAD -> master (permission denied)
error: failed to push some refs to 'github.com:Ryan4253/OkapiLib.git'
```

jazonshou:
> To github.com:Ryan4253/OkapiLib.git

That's odd. I'm trying to merge changes from Yessir120/OkapiLib. Is it possible that there is a problem on my side?

Octogonapus:
I'm not convinced the PR is set up correctly, but regardless, it's now just faster for me to request changes instead

include/okapi/api/chassis/controller/chassisController.hpp
```diff
@@ -57,6 +57,7 @@ class ChassisController {
 
   /**
    * Turns the robot clockwise in place (using closed-loop control).
+   * (turns the shortest angle for OdomChassisControllers)
```
Octogonapus: This isn't applicable to all chassis controllers, so it shouldn't be documented here. You can put it [here](https://github.com/OkapiLib/OkapiLib/blob/master/include/okapi/api/chassis/controller/defaultOdomChassisController.hpp) though. Also please rephrase it to be clearer, e.g. "Turns in the direction of the smallest angle"

include/okapi/api/chassis/controller/chassisController.hpp
```diff
@@ -71,6 +72,7 @@ class ChassisController {
 
   /**
    * Sets the target angle for the robot to turn clockwise in place (using closed-loop control).
+   * (turns the shortest angle for OdomChassisControllers)
```
Octogonapus: same as above

Octogonapus (changes requested):
Please also add a test. Like this one:
```
diff --git a/test/defaultOdomChassisControllerTest.cpp b/test/defaultOdomChassisControllerTest.cpp
index f4de0276..faf0a8f9 100644
--- a/test/defaultOdomChassisControllerTest.cpp
+++ b/test/defaultOdomChassisControllerTest.cpp
@@ -158,6 +158,11 @@ TEST_F(DefaultOdomChassisControllerTest, TurnAboveThreshold) {
   EXPECT_EQ(controller->lastTurnAngleTargetQAngle, 6_deg);
 }
 
+TEST_F(DefaultOdomChassisControllerTest, BoundTurn) {
+  drive->turnToAngle(270_deg);
+  EXPECT_EQ(controller->lastTurnAngleTargetQAngle, -90_deg);
+}
+
 TEST_F(DefaultOdomChassisControllerTest, SetStateTest) {
   auto stateBefore = drive->getState();
   assertOdomStateEquals(0, 0, 0, stateBefore);
```

Octogonapus (approved):

PR #486, opened 2022-06-03
Add curvature drive

Ryan4253:
### Description of the Change

Adds curvature drive (aka cheesy drive) into OkapiLib. It is similar to arcade, but instead of having the right stick control the rate of heading change, it controls the curvature of the path the robot drives in. This allows the robot to turn at a constant rate regardless of how fast your robot is moving, which allows better control under high speeds. A quick demo can be found [here](https://www.youtube.com/watch?v=oUWnYTcbHEw).

The implementation is mostly adapted from [wpilib](https://github.com/wpilibsuite/allwpilib/blob/main/wpilibc/src/main/native/cpp/drive/DifferentialDrive.cpp#L49) (DifferentialDrive.cpp)

### Motivation

From our team's year long usage curvature drive is very nice to drive with especially with high speed chassis. I think a lot of teams will find it useful.

### Possible Drawbacks

Not that I can think of.

### Verification Process

Unit tests were added and passed.
The same code were used on our robot last year and no issues occurred.

### Applicable Issues

None, this is a new feature

### DIFF
--- a/src/api/chassis/model/hDriveModel.cpp
+++ b/src/api/chassis/model/hDriveModel.cpp
@@ -36,7 +36,7 @@ void HDriveModel::forward(const double ispeed) {
 
 void HDriveModel::driveVector(const double iforwardSpeed, const double iyaw) {
   // This code is taken from WPIlib. All credit goes to them. Link:
-  // https://github.com/wpilibsuite/allwpilib/blob/master/wpilibc/src/main/native/cpp/Drive/DifferentialDrive.cpp#L73
+  // https://github.com/wpilibsuite/allwpilib/blob/96e9a6989ce1688f3edb2d9b9d21ef8cd3861579/wpilibc/src/main/native/cpp/Drive/DifferentialDrive.cpp#L73
   const double forwardSpeed = std::clamp(iforwardSpeed, -1.0, 1.0);
   const double yaw = std::clamp(iyaw, -1.0, 1.0);
 
@@ -55,7 +55,7 @@ void HDriveModel::driveVector(const double iforwardSpeed, const double iyaw) {
 
 void HDriveModel::driveVectorVoltage(const double iforwardSpeed, const double iyaw) {
   // This code is taken from WPIlib. All credit goes to them. Link:
-  // https://github.com/wpilibsuite/allwpilib/blob/master/wpilibc/src/main/native/cpp/Drive/DifferentialDrive.cpp#L73
+  // https://github.com/wpilibsuite/allwpilib/blob/96e9a6989ce1688f3edb2d9b9d21ef8cd3861579/wpilibc/src/main/native/cpp/Drive/DifferentialDrive.cpp#L73
   const double forwardSpeed = std::clamp(iforwardSpeed, -1.0, 1.0);
   const double yaw = std::clamp(iyaw, -1.0, 1.0);
 
@@ -87,7 +87,7 @@ void HDriveModel::stop() {
 
 void HDriveModel::tank(const double ileftSpeed, const double irightSpeed, const double ithreshold) {
   // This code is taken from WPIlib. All credit goes to them. Link:
-  // https://github.com/wpilibsuite/allwpilib/blob/master/wpilibc/src/main/native/cpp/Drive/DifferentialDrive.cpp#L73
+  // https://github.com/wpilibsuite/allwpilib/blob/96e9a6989ce1688f3edb2d9b9d21ef8cd3861579/wpilibc/src/main/native/cpp/Drive/DifferentialDrive.cpp#L198
   double leftSpeed = std::clamp(ileftSpeed, -1.0, 1.0);
   if (std::abs(leftSpeed) < ithreshold) {
     leftSpeed = 0;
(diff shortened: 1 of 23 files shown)


src/api/chassis/model/hDriveModel.cpp
```diff
@@ -144,6 +144,42 @@ void HDriveModel::arcade(const double iforwardSpeed, const double iyaw, const do
   middleMotor->moveVelocity(0);
 }
 
+void HDriveModel::curvature(const double iforwardSpeed, 
+                            const double icurvature, 
+                            const double ithreshold) {
+  // This code is adapted from WPIlib. All credit goes to them. Link:
+  // https://github.com/wpilibsuite/allwpilib/blob/main/wpilibc/src/main/native/cpp/drive/DifferentialDrive.cpp#L49
```
baylessj: This should probably get updated to a permalink like https://github.com/wpilibsuite/allwpilib/blob/af7985e46cc3fd6eb01c1243c7f1f95b92ea5ddc/wpilibc/src/main/native/cpp/drive/DifferentialDrive.cpp#L49 in case the WPILib file changes and the current link points to a different part of the code.

src/api/chassis/model/hDriveModel.cpp
```diff
@@ -144,6 +144,42 @@ void HDriveModel::arcade(const double iforwardSpeed, const double iyaw, const do
...
+    forwardSpeed = 0;
+  }
+
+  double curvature = std::clamp(icurvature, -1.0, 1.0);
+  if (std::abs(curvature) < ithreshold) {
+    curvature = 0;
+  }
+
+  // the algorithm switches to arcade when forward speed is 0 to allow point turns.
+  if(forwardSpeed == 0){
```
baylessj: Looks like there are a few code style issues that would be resolved by running `clang-format` per https://github.com/OkapiLib/OkapiLib/blob/master/CONTRIBUTING.md#clang-format

src/api/chassis/model/hDriveModel.cpp
```diff
@@ -144,6 +144,42 @@ void HDriveModel::arcade(const double iforwardSpeed, const double iyaw, const do
...
+
+  // normalizes output
+  if(maxSpeed > 1.0){
+    leftSpeed /= maxSpeed;
+	  rightSpeed /= maxSpeed;
+  }
+
+  leftSideMotor->moveVoltage(static_cast<int16_t>(leftSpeed * maxVoltage));
+  rightSideMotor->moveVoltage(static_cast<int16_t>(rightSpeed * maxVoltage));
+  middleMotor->moveVelocity(0);
```
baylessj, line 180: I'm not entirely sure how often this would be used, but for consistency's sake it might be nice to add an `hCurvature` function like how `arcade` has a `hArcade` companion.

baylessj (changes requested):
Added a few requests for cleanup to this PR.

I don't intend this as a requirement for completing this PR or anything, but I'm curious about your opinions on difference in performance between WPILib's curvature drive here and 254's Cheesy Drive (an example implementation for VEX here: https://github.com/purduesigbots/forkner-public/blob/62e1328b7902715035357622d81e4a35cb15ff2f/src/libforkner/drive.cpp#L303). Cheesy Drive is quite a bit more complicated, but I found it to drive really nicely when I competed. I never gave WPILib's more basic curvature drive a shot though, so I'd be interested in your feedback if you could compare the two at some point.

src/api/chassis/model/hDriveModel.cpp
Ryan4253: I have added the hCurvature() implementation in the latest commit. Thank you.

I was also thinking about whether to implement an xCurvature() function for the sake of consistency. However, I don't think scaling the (independent) rotational output with the forward output would make much sense for a x drive. I know that @Yessir120 has a x drive on hand right now, and maybe he can help test if the algorithm would make any sense.

jazonshou: > I don't think scaling the (independent) rotational output with the forward output would make much sense for a x drive

Agree. I don't think it would make sense, especially when strafing gets added.

jazonshou:
> WPILib's curvature drive here and 254's Cheesy Drive

Yeah, I agree, they certainly are different. I'm not 100% certain about this, but I'm pretty sure cheesy drive first originated as the current WPILib implementation of the curvature drive. Then, as seasons progressed, 254 added extra features like slew controllers to make driving smoother. 

I think it could be interesting to see how much improvement the extra features from cheesy drive make. But being someone who has both driven VEX and FRC robots, I will say that FRC robots are much harder to control. So there is a possibility that cheesy drive might be too much for VEX robots.

src/api/chassis/model/hDriveModel.cpp
Ryan4253: The link have been replaced. I also went ahead and replaced the links in functions like tank() and driveVector() because they are all dead.

src/api/chassis/model/hDriveModel.cpp
Ryan4253: I have a quick question on how to approach this. I have ran clang-format, but it seems like there were other files (such as asyncMotionProfileController) that were formatted as well.
image

Should I only commit the files that I've edited, or should I go ahead and commit all the files that were formatted.

Ryan4253: Well, I have gone ahead and modified all of them in the latest commit. The formatting issues should be fixed now.

baylessj (approved):
The changes look good to me! Looks like there were a number of places in the project that hadn't gotten clang-format run on them recently... not sure if it matters that those changes are being added in this PR vs. a dedicated formatting one.

PR #487, opened 2022-06-10
Allow varying units in printing odomState

### LINKED ISSUE #467: Allow users to set units for OdomState string representation
When logging odometry states, we can use the OdomState.str() function to do so easily. However, the function can only log the value in meters, even though it's usually easier to think in terms of feet. Although it's easy to do a rough conversion in your head, it's annoying to determine exact values.

Parameters in the OdomState.str() function to pass in the unit (e.g. feet, inches, etc). The new function would look like this:
```cpp
std::string OdomState::str(QLength idistanceUnit=meter, QAngle iangleUnit=degree) const {
  std::ostringstream os;
  os << "OdomState(x=" << std::to_string(x.convert(idistanceUnit))
     << ", y=" << std::to_string(y.convert(idistanceUnit))
     << ", theta=" << std::to_string(theta.convert(iangleUnit)) << ")";
  return os.str();
}
```

The problem with this solution is that the unit is no longer listed in the string representation. I'm not sure if there's an easy way to convert a unit to a string representation already.

Users can create the string themselves if they'd like to use different units.

**Versions**
PROS: 3.3.1
Okapi: 4.2.0

karmanyaahm:
Test this with:
```cpp
void opcontrol()
{
        using namespace okapi::literals;
        auto o = okapi::OdomState{x : 4.20_tile, y : 31.4_in, theta : 69_deg};

        try
        {
                printf("%s\n", o.str().c_str()); //normal, compatible with existing behavior
                printf("%s\n", o.str(1_ft, 1_deg).c_str()); // basic unit specification
                printf("%s\n", o.str(okapi::meter, okapi::radian).c_str()); // another basic unit specification
                printf("%s\n", o.str(6_tile / 100, "%", 360_deg / 100, "%").c_str()); // fancy custom unit specification (% of field)
                printf("%s\n", o.str(1_m, 1.01_deg)); // should fail because no matching unit
        }
        catch (const std::domain_error &e)
        {
                printf("%s\n", e.what());
        }
}
```

TODO:
- [x] docs
- [x] Rebase: clean up commit history

Closes #467

### DIFF
--- a/include/okapi/api/units/RQuantityName.hpp
+++ b/include/okapi/api/units/RQuantityName.hpp
@@ -9,6 +9,14 @@
 
 namespace okapi {
 
+/**
+* Returns a short name for a unit.
+* For example: `str(1_ft)` will return "ft", so will `1 * foot` or `0.3048_m`.
+* Throws std::domain_error when `q` is a unit not defined in this function.
+*
+* @param q Your unit. Currently only QLength and QAngle are supported.
+* @return The short string suffix for that unit.
+*/
 template <class QType> std::string getShortUnitName(QType q) {
   const std::unordered_map<std::type_index, std::unordered_map<double, const char *>> shortNameMap =
     {{typeid(meter),
--- a/include/okapi/api/units/RQuantityName.hpp
+++ b/include/okapi/api/units/RQuantityName.hpp
@@ -1,0 +1,38 @@
+#include "okapi/api/units/QAngle.hpp"
+#include "okapi/api/units/QLength.hpp"
+#include "okapi/api/units/QSpeed.hpp"
+#include <stdexcept>
+#include <typeindex>
+#include <unordered_map>
+
+#pragma once
+
+namespace okapi {
+
+template <class QType> std::string getShortUnitName(QType q) {
+  const std::unordered_map<std::type_index, std::unordered_map<double, const char *>> shortNameMap =
+    {{typeid(meter),
+      {
+        {meter.getValue(), "m"},
+        {decimeter.getValue(), "dm"},
+        {centimeter.getValue(), "cm"},
+        {millimeter.getValue(), "mm"},
+        {kilometer.getValue(), "km"},
+        {inch.getValue(), "in"},
+        {foot.getValue(), "ft"},
+        {yard.getValue(), "yd"},
+        {mile.getValue(), "mi"},
+        {tile.getValue(), "tile"},
+      }},
+     {typeid(degree), {{degree.getValue(), "deg"}, {radian.getValue(), "rad"}}}};
+
+  try {
+    return shortNameMap.at(typeid(q)).at(q.getValue());
+  } catch (const std::out_of_range &e) {
+    throw std::domain_error(
+      "You have requested the shortname of an unknown unit somewhere (likely odometry strings). "
+      "Shortname for provided unit is unspecified. You can override this function to add more "
+      "names or manually specify the name instead.");
+  }
+}
+} // namespace okapi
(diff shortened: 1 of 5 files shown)


include/okapi/api/units/RQuantityName.hpp
```diff
@@ -0,0 +1,37 @@
...
+       {kilometer.getValue(), "km"},
+       {inch.getValue(), "in"},
+       {foot.getValue(), "ft"},
+       {yard.getValue(), "yd"},
+       {mile.getValue(), "mi"},
+       {tile.getValue(), "tile"},
+      }},
+     {typeid(degree), {{degree.getValue(), "deg"}, {radian.getValue(), "rad"}}}};
+
+  try {
```
karmanyaahm, line 37: Should I make a cpp file for this, or should I leave this in hpp?
If making a cpp, should the map be in cpp or hpp?

src/api/odometry/odomState.cpp
```diff
@@ -15,11 +15,20 @@ bool OdomState::operator!=(const OdomState &rhs) const {
...
-  std::ostringstream os;
-  os << "OdomState(x=" << std::to_string(x.convert(meter))
-     << "m, y=" << std::to_string(y.convert(meter))
-     << "m, theta=" << std::to_string(theta.convert(degree)) << "deg)";
-  return os.str();
+std::string OdomState::str(QLength idistanceUnit,
+                           std::string distUnitName,
+                           QAngle iangleUnit,
+                           std::string angleUnitName) const {
+  char buf[150];
```
karmanyaahm, line 22: Rewrote because the '<<' syntax was getting messy when the units were also specified. Is this fine? 
A 'normal' string `OdomState(x=2.56m, y=0.80m, theta=69.00deg)` is 44 bytes, so 150 ought to be more than enough for anything.

This is using snprintf because PROS uses c++17 by default, which doesn't support std::format. https://github.com/purduesigbots/pros/issues/375

karmanyaahm:
@Octogonapus Can you please answer the questions in the review comments?

src/api/odometry/odomState.cpp
```diff
@@ -15,11 +15,20 @@ bool OdomState::operator!=(const OdomState &rhs) const {
...
+  snprintf(buf, sizeof(buf), "OdomState(x=%.2f%s, y=%.2f%s, theta=%.2f%s)", x.convert(idistanceUnit), distUnitName.c_str(), y.convert(idistanceUnit), distUnitName.c_str(), theta.convert(iangleUnit), angleUnitName.c_str());
+  return std::string(buf);
```
karmanyaahm: Rewrote because the '<<' syntax was getting messy when the units were also specified. Is this fine? 
A 'normal' string `OdomState(x=2.56m, y=0.80m, theta=69.00deg)` is 44 bytes, so 150 ought to be more than enough for anything.

This is using snprintf because PROS uses c++17 by default, which doesn't support std::format. https://github.com/purduesigbots/pros/issues/375

include/okapi/api/units/RQuantityName.hpp
Octogonapus: No, this is fine.

src/api/odometry/odomState.cpp
Octogonapus: This is fine.

karmanyaahm:
@Octogonapus can you look at this please?

Octogonapus (approved):
pretty sure you need double backticks for inline code but we'll see

Octogonapus:
update: no that actually works

